PR #3, opened 2019-10-25
Permissions new

GuillaumeCisco:


thibaultrobert:
Is this still an active PR since the permissions have been merged and the `dev` branch has changed significantly since then? May I close it (we can keep the branch if needed)?

GuillaumeCisco:
I think this can be closed now

PR #6, opened 2019-10-30
Fix staticcheck warnings

olivierperes:
These warnings were reported by [staticcheck](https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck).

### DIFF
--- a/chaincode/permissions.go
+++ b/chaincode/permissions.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 )
 
+/* Unused
 // Action is the the type of an action
 type Action string
 
@@ -26,6 +27,7 @@ const (
 	Process  Action = "process"
 	Download Action = "download"
 )
+*/
 
 // Permission represents one permission based on an action type
 type Permission struct {
@@ -82,7 +84,7 @@ func NewPermissions(db LedgerDB, in inputPermissions) (Permissions, error) {
 		}
 
 		if !stringInSlice(authorizedID, nodesIDs) {
-			return Permissions{}, fmt.Errorf("Invalid permission input values")
+			return Permissions{}, fmt.Errorf("invalid permission input values")
 		}
 	}
 
@@ -105,10 +107,7 @@ func newPermission(in inputPermission, owner string) Permission {
 	if !stringInSlice(owner, in.AuthorizedIDs) {
 		in.AuthorizedIDs = append([]string{owner}, in.AuthorizedIDs...)
 	}
-	return Permission{
-		Public:        in.Public,
-		AuthorizedIDs: in.AuthorizedIDs,
-	}
+	return Permission(in)
 }
 
 func (priv Permission) include(other Permission) bool {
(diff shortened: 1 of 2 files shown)


chaincode/input.go
```diff
@@ -162,14 +162,14 @@ type inputComputePlanTesttuple struct {
 
 type inputLeaderboard struct {
 	ObjectiveKey   string `validate:"omitempty,len=64,hexadecimal" json:"objectiveKey"`
-	AscendingOrder bool   `json:"ascendingOrder,required"`
+	AscendingOrder bool   `validate:"required" json:"ascendingOrder"`
```
thibaultrobert: Be careful the `required` tag  in the lib validator.v9 checks if the field is not zero value : https://godoc.org/gopkg.in/go-playground/validator.v9#hdr-Required
So for a boolean it will enforce the `True` value whereas the json `required` tag will just ensure one value is passed, either `False` or `True`. So let's keep the json one for now.

chaincode/input.go
```diff
@@ -162,14 +162,14 @@ type inputComputePlanTesttuple struct {
...
 }
 
 type inputPermissions struct {
 	Process inputPermission `validate:"required" json:"process"`
 }
 
 type inputPermission struct {
-	Public        bool     `json:"public,required"`
+	Public        bool     `validate:"required" json:"public"`
```
thibaultrobert: Same here.

chaincode/permissions.go
```diff
@@ -26,14 +27,15 @@ const (
...
+*/
 
 // Permission represents one permission based on an action type
 type Permission struct {
 	// Public is true if this permission is given to the asset's owner only and
 	// the nodes listed in AuthorizedIDs (open to all nodes if false)
 	Public bool `json:"public"`
 	// AuthorizedIDs list all authorised nodes other than the asset's owner
-	AuthorizedIDs []string `json:"authorizedIDs", omitempty`
+	AuthorizedIDs []string `json:"authorizedIDs,omitempty"`
```
thibaultrobert: This change is a nice catch but it belongs in its own PR and requires to update the tests since it breaks some of them.

thibaultrobert (changes requested):
Although some changes are nice catch, they are too much side-effects to this PR. Moreover it breaks the tests.

chaincode/input.go
olivierperes: OK, removed `validate:"required"` in both instances.

Also left `json:required` because removing it breaks the tests, although I don’t see that it does anything. What am I missing?

chaincode/permissions.go
olivierperes: OK, removed. I will submit a separate PR.

chaincode/input.go
olivierperes: Done, as above.

Kelvin-M:
@olivierperes Can you rebase on `dev` to be able to do the review ?
@thibaultrobert  After the rebase, it seems to be good to be merged ?

olivierperes:
@Kelvin-M Rebased.

Kelvin-M:
Thank you, I let @thibaultrobert complete his review to see if some extra work has to be done :)

thibaultrobert (approved):
Thanks a lot for the contribution. I think the `Unused` sections can be removed and I invite you to do so in another PR, if you're willing of course.

olivierperes:
> Thanks a lot for the contribution. I think the `Unused` sections can be removed and I invite you to do so in another PR, if you're willing of course.

Sure, done in #35.

PR #13, opened 2019-10-30
Refactoring/split tuple

thibaultrobert:
Since we intend to add new tuple types, we need to split `tuple.go` and its tests into different files


chaincode/traintuple.go
```diff
@@ -0,0 +1,624 @@
...
+			out := outputTesttuple{}
+			err = out.Fill(db, testtupleKey, testtuple)
+			if err != nil {
+				return err
+			}
+			event.AddTesttuple(out)
+		}
+	}
+	return nil
+}
```
inelgnu: missing EOF

inelgnu (dismissed):

chaincode/traintuple.go
thibaultrobert: Thanks and done.

inelgnu (approved):

PR #14, opened 2019-10-30
Fix invalid JSON tag

olivierperes:
Reported by [staticcheck](https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck).

If `omitempty` is intended, this requires fixing `TestAlgo` in a way that is unclear to me. I am not sure that `testify/assert` supports it. In that case, I could open an issue instead and close the PR as it is probably not my place to introduce such a change in a test.

### DIFF
--- a/chaincode/permissions.go
+++ b/chaincode/permissions.go
@@ -33,7 +33,7 @@ type Permission struct {
 	// the nodes listed in AuthorizedIDs (open to all nodes if false)
 	Public bool `json:"public"`
 	// AuthorizedIDs list all authorised nodes other than the asset's owner
-	AuthorizedIDs []string `json:"authorizedIDs", omitempty`
+	AuthorizedIDs []string `json:"authorizedIDs"`
 }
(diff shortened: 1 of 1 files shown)


Kelvin-M:
Thank you for your contribution @olivierperes 
In `substra`, the permission design allows the user to provide an empty `AuthorizedIDs`, so we need to keep the `omitempty`.
Another PR (#17) found a potential issue, not directly link with yours, but suggested that we need to review the permission code.
I will let @thibaultrobert or @inalgnu close this PR if not relevant and create a issue or PR to fix also this invalid tag :)

thibaultrobert:
You're right about this tag it is faulty. But as @Kelvin-M said the `omitempty` should be present and the tests pass with it correctly set. But this change does required a bit of in depth comprehension of the tests. I'll open an issue on this topic. Please feel free to fix it if you want or to contribute otherwise to Substra. Thanks again.

PR #15, opened 2019-10-31
readme: add section about local development

AurelienGasser:
It's the second time I spend 20+ minutes looking for which go version HLF uses. It's [not very easy to find](https://hyperledger-fabric.readthedocs.io/en/release-1.4/prereqs.html#go-programming-language) a definitive answer :)

Hopefully this section can be built upon in the future!

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,20 @@ Chaincode for the Substra platform
 ## License
 
 This project is developed under the Apache License, Version 2.0 (Apache-2.0), located in the [LICENSE](./LICENSE) file.
+
+## Local development
+
+### Prerequisites
+
+go version 1.11.x
+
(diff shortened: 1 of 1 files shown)


AurelienGasser:
Please also note that the chaincode [currently uses go version 1.12](https://github.com/SubstraFoundation/substra-chaincode/blob/dev/chaincode/go.mod#L3) (and not 1.11.x), though I don't think it's a big deal.

thibaultrobert:
It might in fact be problematic the more we rely on modules since it's what ensures backward compatibility. We should at least test it in the `1.11` in the CI and update the `go.mod`.

thibaultrobert (approved):

PR #17, opened 2019-11-04
Add tests on permissions

GuillaumeCisco:
I was looking at the code of the permissions and found some incoherence I'd like to share.

Currently the implementation of a permission is:
```go
// Permission represents one permission based on an action type
type Permission struct {
	// Public is true if this permission is given to the asset's owner only and
	// the nodes listed in AuthorizedIDs (open to all nodes if false)
	Public bool `json:"public"`
	// AuthorizedIDs list all authorised nodes other than the asset's owner
	AuthorizedIDs []string `json:"authorizedIDs, omitempty"`
}
```

So we can have at least 4 possibilities when giving inputs.

#### The Public one:
```
{
    public: true,
    authorizedIds: []
}
```

#### The Private one:
```
{
    public: false,
    authorizedIds: ["foo", "bar"]
}
```

#### The Private one with only owner (will be set implicitely):
```
{
    public: false,
    authorizedIds: []
}
```

#### The Bad input one:
```
{
    public: true,
    authorizedIds: ["foo", "bar"]
}
```

The last one is problematic, as from a user point of view, it looks like we only restrict the permissions to `foo`, `bar` and ourself.

__`public: true` is a miss.__

This does not feel good.  
The code should raise an error but after writing the test, I had no error.  
The permission is:
```
{
    public: true,
    authorizedIds: ["owner", "foo", "bar"]
}
```
which does not feel right.

---------------

Furthermore I've seen this comment in the code on `newPermission` function:
```
// Owner must always be defined in the list of authorizedIDs, if the permission is private,
// it will ease the merge of private permissions
```
Unfortunately, even if the permission is public, the owner is set in the `authorizedIDs` parameter.
I've written tests about that too.

---------------

In a nutshell, using an object for permission can be very cumbersome.  
Simply using an nullable slice seems a good solution, we would have:

#### The Public one:
```
null
```

#### The Private one:
```
["foo", "bar"]
```

#### The Private one with only owner (will be set implicitely):
```
[]
```


Kelvin-M:
Nice catch ! 
We should first fix, with the current permission definition, those two corner cases. @thibaultrobert @inalgnu ! It may have side effects on the backend (I may need to take time to check the behaviour in substrabac in those two cases)

Changing the structure of the permissions will have an impact on multiple repo so we may need to do a mission for that.
I don't know if changing permission as suggested will be an improvement for the future but there is place for improvement as it seems that `public` state + `authorizedIds ` are maybe redudant

Kelvin-M:
@GuillaumeCisco This branch should be rebased :)
@thibaultrobert Do you have an opinion on this issue ?

GuillaumeCisco:
Just rebased ;)

Kelvin-M:
Close it as no comment was made since nearly a year

PR #21, opened 2019-11-05
Testtuples: reference the traintuple instead of referencing the model directly

AurelienGasser:
See also: https://github.com/SubstraFoundation/substra-backend/pull/33

---

Store a reference to the traintuple instead of a reference to the model. The model can still be obtained by querying the traintuple. This change will make it easier to implement generic test tuples.

In this PR:

- Add a new field `traintupleKey` to `Testtuple`, `outputTesttuple` and `outputBoardTuple`
- Delete the `model` field from `Testtuple`
- Mark the `model` field from `outputTesttuple` and `outputBoardTuple`
  - We don't delete these fields yet to prevent potential breakages in other parts of the substra stack

Future work:

- Delete the `model` field from `outputTesttuple` and `outputBoardTuple`


chaincode/output.go
```diff
@@ -199,16 +199,17 @@ func (outputTraintuple *outputTraintuple) Fill(db LedgerDB, traintuple Traintupl
...
-	Status    string         `json:"status"`
-	Tag       string         `json:"tag"`
+	Key           string         `json:"key"`
+	Algo          *HashDressName `json:"algo"`
+	Certified     bool           `json:"certified"`
+	Creator       string         `json:"creator"`
+	Dataset       *TtDataset     `json:"dataset"`
+	Log           string         `json:"log"`
+	TraintupleKey string         `json:"traintupleKey"`
+	Model         *Model         `json:"model"` // Obsolete field. TODO: delete it
```
Kelvin-M, line 209: If it's obsolete, it should be removed no ?

Kelvin-M (commented):
I notice that the testtuple is changing and we need to do an extra call to get the model hash, is this something that we want ?
Moreover, I think it will break some stuff on the frontend (because the backend is changing too)

chaincode/output.go
thibaultrobert: It was but put back to guarantee backward compatibility. But as far as we know, nothing breaks if it's removed (with the PR https://github.com/SubstraFoundation/substra-backend/pull/33). We intend to do so in a future PR.

Kelvin-M: @thibaultrobert We do backward compatibility on the chaincode ?
I thought we need a upgrade function before dealing with backward compatibility on the chaincode, no ?

thibaultrobert: You right about that. We do not support chaincode update for now. Sorry if my statement was misleading.
The backward compatibility is not between two chaincode version but between the two different we deal with models, the ones before and after this PR. As I say the PR https://github.com/SubstraFoundation/substra-backend/pull/33 should ensure that deleting the `model` field won't have side effects but we wish to invest more time be sure. So we keep both approach for now and we'll come back to it later.

Kelvin-M: Okay ! Thank you for those information. I put also a global comment on my review that it may have an impact (I did not test it) on substra-frontend :)

AurelienGasser:
@Kelvin-M Regarding the extra ledger call, please see [the conversation there](https://github.com/SubstraFoundation/substra-backend/pull/33#discussion_r343514397)

Regarding the frontend breaking, I've tested on my local: I generated all kinds of assets and clicked around on _everything_. I've not managed to see anything broken :) All HTTP calls go through, and no error in the console. I could have missed something though, and I think it would be worth having @jmorel 's opinon.

Kelvin-M:
@AurelienGasser Nice ! Did you test it also with the removal of the `Model` key/value ?

AurelienGasser:
@Kelvin-M I did :)

Kelvin-M:
Okay, we wait for jmorel to check if everything is good :)

Kelvin-M (dismissed):
I've tested it, it works ! Nice work :)
As suggested it could be nice to remove the `Model` of `outputTesttuple` as it works without it.

jmorel:
About the frontend, we always use the traintuple to get model details, we never used the testtuple to do so. I think it's safe to remove from the frontend points of view.

Kelvin-M:
So we can remove `Model` in this PR :)
What do you think @AurelienGasser  & @thibaultrobert  ?

AurelienGasser:
@Kelvin-M No objection, other than @inalgnu recommended against it (although arguably that was before you and @jmorel double checked that the field is not used!)

Kelvin-M (approved):

Kelvin-M:
@thibaultrobert Need you review for this to be merged !

thibaultrobert (approved):
LGTM and tested

PR #23, opened 2019-11-07
downgrade go version from 1.12 to 1.11

### LINKED ISSUE #20: Update go.mod to go version 1.1
See https://github.com/SubstraFoundation/substra-chaincode/pull/15

> We should at least test it in the 1.11 in the CI and update the go.mod.

Currently is uses v 1.12, which is not officially supported by HLF

AurelienGasser:
Fixes https://github.com/SubstraFoundation/substra-chaincode/issues/20


thibaultrobert:
I've commit an update to be coherent with the go version specified in the module file. The methods `ReplaceAll` from the package `strings` is not compatible with go `1.11` ([See here](https://golang.org/pkg/strings/#ReplaceAll)). I've also update the modules files with a 'go mod tidy' command to remove unwanted dependencies.

thibaultrobert (approved):

PR #34, opened 2019-11-15
Allow traintuples with the same compute plan id to be generated from different algos

samlesu:
Currently this is not possible: an error is raised when creating a traintuple if its algo is different from the traintuples algo of this compute plan.

Removing this check could allow to create multiple traintuples from different algos to share the same local folder. 

This could be the first step before creating a compute plan with traintuples of different algos.

Tests should be updated, but I'm first opening this PR to discuss about it.

### DIFF
--- a/chaincode/traintuple.go
+++ b/chaincode/traintuple.go
@@ -155,16 +155,6 @@ func (traintuple *Traintuple) AddToComputePlan(db LedgerDB, inp inputTraintuple,
 	if len(ttKeys) == 0 {
 		return errors.BadRequest("cannot find the ComputePlanID %s", inp.ComputePlanID)
 	}
-	for _, ttKey := range ttKeys {
-		FLTraintuple, err := db.GetTraintuple(ttKey)
-		if err != nil {
-			return err
-		}
-		if FLTraintuple.AlgoKey != inp.AlgoKey {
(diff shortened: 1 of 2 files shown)


thibaultrobert:
It's a good and necessary step. It would have been done in the future PR on compute plan for composite traintuple anyway since traintuple and composite traintuple can't have the same algo. Do you think it can wait for the composite PR or you want to fix the tests and merge this PR instead?

samlesu:
Thanks for your answer!

I would prefer to not wait as these PRs are independent and we don't know all long we will have to wait :)

AurelienGasser:
What about adding a unit test with the "hybrid algo compute plan" scenario ?

AurelienGasser:
Wow, that test implementation was fast :laughing:

samlesu:
> What about adding a unit test with the "hybrid algo compute plan" scenario ?


It looks like the test already exists, just need to update it.

AurelienGasser (approved):
LGTM! :+1:

samlesu:
Note: once this PR is merged, we should create an issue to update the CreateComputePlan smart contract in order to support multiple algos.

samlesu:
@inalgnu @thibaultrobert Could you please review this PR? Thanks :)

inelgnu (approved):

PR #41, opened 2019-11-26
Compute plan: support composite/aggregate

### LINKED ISSUE #31: [BUG] queryComputePlans returns null instead of [] when there are no computePlans
**Describe the bug**

All "queryAssets" functions return an empty list when there are no assets available. `queryComputePlans` on the other hand returns `null`. 

**To Reproduce**

* call queryComputePlans on a brand new instance of the platform

**Expected behavior**

Expected return value: `[]`

Actual return value: `null`

### LINKED ISSUE #36: [BUG] createComputePlan and queryComputePlan(s) have different outputs
**Describe the bug**
Use `createComputePlan` and you'll get an `outputComputePlan` while the `queryComputePlan` and `queryComputePlans` method return `outputComputePlanDetails`

This means that the former one doesn't include the `algoKey` and `objectiveKey` while the latter ones use `traintuples`/`testtuples` instead of `traintupleKeys`/`testtupleKeys`.

**Expected behavior**

They should return the same data in the same format.

This common format could be a mix of the two structs with the attributes:
* `computePlanID`
* `algoKey`
* `objectiveKey`
* `traintupleKeys`
* `testtupleKeys`

I offer to use traintupleKeys and testtupleKeys because we don't include the full traintuples and testtuples in the compute plan.

Also, when the testtupleKeys is empty, an empty list should be returned (instead of `null`)

AurelienGasser:
Companion PR: [substra-backend](https://github.com/SubstraFoundation/substra-backend/pull/57)

This updates the compute plan structures to support composite/aggregate.

Fixes https://github.com/SubstraFoundation/substra-chaincode/issues/31

Fixes https://github.com/SubstraFoundation/substra-chaincode/issues/36

### DIFF
--- a/chaincode/compute_plan_dag.go
+++ b/chaincode/compute_plan_dag.go
@@ -2,6 +2,8 @@ package main
 
 import "fmt"
 
+// TrainingTask is a node of a ComputeDAG. It represents a training task
+// (i.e. a Traintuple, a CompositeTraintuple or an Aggregatetuple)
 type TrainingTask struct {
 	ID          string
 	InModelsIDs []string
@@ -9,10 +11,13 @@ type TrainingTask struct {
 	TaskType    AssetType
 }
 
+// ComputeDAG is a Directed Acyclic Graph (DAG)
+// used for compute plans
 type ComputeDAG struct {
 	OrderTasks []TrainingTask
 }
 
+// Create a Directed Acyclic Graph (DAG) from a compute plan
 func createComputeDAG(cp inputComputePlan) (ComputeDAG, error) {
 	DAG := ComputeDAG{}
 	for i, traintuple := range cp.Traintuples {
@@ -49,8 +54,7 @@ func createComputeDAG(cp inputComputePlan) (ComputeDAG, error) {
 	return DAG, nil
 }
 
-// sort order the listed task of the dag or return an error if there is a cyclic
-// dependencies in the inModelIDs
+// Sort the DAG's task list, or return an error if there is a cyclic dependency in inModelIDs
 func (dag *ComputeDAG) sort() error {
 	current := dag.OrderTasks
 	var temp, final []TrainingTask
@@ -67,7 +71,7 @@ func (dag *ComputeDAG) sort() error {
 		if ready {
 			final = append(final, current[i])
 			if _, ok := IDPresents[current[i].ID]; ok {
-				return fmt.Errorf("compute plan error, ID use twice: %s", current[i].ID)
+				return fmt.Errorf("Compute plan error: Duplicate training task ID: %s", current[i].ID)
 			}
 			IDPresents[current[i].ID] = true
 		} else {
@@ -82,7 +86,7 @@ func (dag *ComputeDAG) sort() error {
 			for _, c := range current {
 				errorIDs = append(errorIDs, c.ID)
 			}
-			return fmt.Errorf("compute plan error, either cyclic or missing dep among those IDs'inModels: %v", errorIDs)
+			return fmt.Errorf("Compute plan error: Cyclic or missing dependency among inModels IDs: %v", errorIDs)
 		}
 		i = 0
 		current = temp
--- a/chaincode/compute_plan_dag.go
+++ b/chaincode/compute_plan_dag.go
@@ -71,7 +71,7 @@ func (dag *ComputeDAG) sort() error {
 		if ready {
 			final = append(final, current[i])
 			if _, ok := IDPresents[current[i].ID]; ok {
-				return fmt.Errorf("Compute plan error: Duplicate training task ID: %s", current[i].ID)
+				return fmt.Errorf("compute plan error: Duplicate training task ID: %s", current[i].ID)
 			}
 			IDPresents[current[i].ID] = true
 		} else {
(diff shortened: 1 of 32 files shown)


chaincode/compute_plan_dag.go
```diff
@@ -67,7 +71,7 @@ func (dag *ComputeDAG) sort() error {
 		if ready {
 			final = append(final, current[i])
 			if _, ok := IDPresents[current[i].ID]; ok {
-				return fmt.Errorf("compute plan error, ID use twice: %s", current[i].ID)
+				return fmt.Errorf("Compute plan error: Duplicate training task ID: %s", current[i].ID)
```
thibaultrobert: The error message should not start with a capital letter since it can be composed later on with another.

chaincode/ledger.go
```diff
@@ -88,57 +88,72 @@ type AggregateAlgo struct {
...
 	InHeadModel   string                      `json:"inHeadModel"`
 	InTrunkModel  string                      `json:"inTrunkModel"`
-	Log           string                      `json:"log"`
-	ObjectiveKey  string                      `json:"objectiveKey"`
 	OutHeadModel  CompositeTraintupleOutModel `json:"outHeadModel"`
 	OutTrunkModel CompositeTraintupleOutModel `json:"outTrunkModel"`
 	Perf          float32                     `json:"perf"`
-	Rank          int                         `json:"rank"`
-	Status        string                      `json:"status"`
-	Tag           string                      `json:"tag"`
```
thibaultrobert, line 125: I don't mind resorting struct element but I'm not a sure what's the logic behind this order (also an alphabetic order was ok for me)

thibaultrobert (dismissed):

chaincode/compute_plan_dag.go
AurelienGasser: I see @inalgnu already fixed it! Under what scenario would we compose this error with others?

chaincode/main.go
```diff
@@ -170,6 +170,12 @@ func (t *SubstraChaincode) Invoke(stub shim.ChaincodeStubInterface) peer.Respons
 	if err != nil {
 		return formatErrorResponse(err)
 	}
+	// Send event if there is any. It's done in one batch since we can only send
+	// one event per call
+	err = db.SendTuplesEvent()
+	if err != nil {
+		return formatErrorResponse(fmt.Errorf("could not send event for unknown reason"))
```
AurelienGasser: Is there a reason not to add the error returned by `SendTuplesEvent` as part of this error, instead of saying the error is unknown?

chaincode/traintuple.go
```diff
@@ -377,22 +363,12 @@ func logFailTrain(db LedgerDB, args []string) (outputTraintuple outputTraintuple
...
-	event := TuplesEvent{}
-	err = UpdateTesttupleChildren(db, inp.Key, traintuple.Status, &event)
-	if err != nil {
-		return
-	}
-
-	err = UpdateTraintupleChildren(db, inp.Key, traintuple.Status, &event)
-	if err != nil {
-		return
-	}
```
AurelienGasser, line 375: These err blocks have been removed, is it intentional?

AurelienGasser: nevermind, I just can't read right :D

chaincode/compute_plan_dag.go
inelgnu: @AurelienGasser https://github.com/golang/go/wiki/CodeReviewComments#error-strings

chaincode/ledger_db.go
```diff
@@ -383,3 +384,73 @@ func (db *LedgerDB) GetNode(key string) (Node, error) {
...
 	return node, nil
 }
+
+// ----------------------------------------------
+// High-level functions for events
+// ----------------------------------------------
+
+// SendTuplesEvent sends an event with updated tuples if there is any
+// Only one event can be sent per transaction
+func (db *LedgerDB) SendTuplesEvent() error {
```
AurelienGasser, line 394: I think you forgot to remove the original method from `utils.go`

chaincode/ledger.go
AurelienGasser: I sorted it this way so that the common members between Traintuple, CompositeTraintuple and AggregateTuple are all at the top (and hopefully can eventually be removed altogether, in favor of structure composition where `GenericTuple` would be included as part of each sub-structure? to be discussed :p)

I can revert to the original sort, if you prefer?

chaincode/compute_plan_dag.go
AurelienGasser: Thank you! 

Following that logic, what would you think of removing the "Compute plan error: " part of the error string?

AurelienGasser: Talked with @thibaultrobert : let's leave it as it is now.

chaincode/ledger.go
AurelienGasser: discussed with @thibaultrobert : no strong opinion, let's leave it as it is now.

chaincode/ledger_db.go
AurelienGasser: confirmed with @thibaultrobert this needs fixing

chaincode/main.go
AurelienGasser: confirmed with @thibaultrobert this needs fixing

chaincode/ledger_db.go
AurelienGasser: fixed: https://github.com/SubstraFoundation/substra-chaincode/pull/41/commits/d06ce2fd5387ea3c48400161227abd2f6c1be4d2

chaincode/main.go
AurelienGasser: fixed: https://github.com/SubstraFoundation/substra-chaincode/pull/41/commits/53ccf231c878e305f5aef1e17ca09dc5671690d8

inelgnu (dismissed):

thibaultrobert (approved):

PR #49, opened 2019-12-02
travis: optimize lint

AurelienGasser:
**1. Lint only our modules:**

- Running `golint` doesn't lint the `errors/` package.
- Running `golint ./...` lints the `vendors/` package, which is wrong too.
- => Use `go list` with a grep in order to lint only the packages we wrote

**2. Try building the chaincode even if linting fails**

Rationale: I assume it's always interesting to see if the package can build, even if linting failed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,5 +20,5 @@ script:
   - go test -v ./...
   - golint_stdout="`go list ./... | grep -v /vendor/ | xargs golint`"
   - go build -v  
-  - test -z "$golint_stdout" || (echo -e "golint error(s):\n$golint_stdout" && exit 1)
+  - test -z "$golint_stdout" || (echo -e "FAIL:" "golint error(s):\n$golint_stdout" && exit 1)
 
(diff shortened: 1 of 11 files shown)


AurelienGasser:
Running `golint` _after_ `go build` leads to the error

```
chaincode:1:1: illegal character U+007F
```

I'll try a new approach: lint before the build, but try building even if linting fails.

AurelienGasser:
Looks like it works:

- It caught the error in the main directory
- It caught the error in the `errors/` subdirectory
- It still tried to `go build` the package
- The CI build gets marked as "failed" out because of the lint errors

I'll remove the intentional lint errors.

AurelienGasser:
This is ready for review. 

Sorry for the noise !

AurelienGasser:
Today I learned: adding `: ` (colon followed by a space) to a travis command makes the YAML invalid! Because `: ` is part of the YAML syntax.

thibaultrobert (approved):

PR #52, opened 2019-12-05
add getComputePlanStatus

AurelienGasser:

PR #53, opened 2019-12-06
Cancel compute plan

inelgnu:
Companion PR:
- [substra](https://github.com/SubstraFoundation/substra/pull/56)
- [backend](https://github.com/SubstraFoundation/substra-backend/pull/65)
- [tests](https://github.com/SubstraFoundation/substra-tests/pull/26)
- [frontend](https://github.com/SubstraFoundation/substra-frontend/pull/33)

### DIFF
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -313,3 +313,57 @@ func getComputePlan(db *LedgerDB, key string) (resp outputComputePlan, err error
 	}
 	return
 }
+
+func getComputePlanStatus(db *LedgerDB, args []string) (status string, err error) {
+
+	AssetFromJSON(args)
+
+	computePlan, err := getComputePlan(db, key)
+	if err != nil {
+		return "", err
+	}
+
+	// get all tuples like status in one slice
+	statusCollection := []string{}
+
+	for _, aggregatetupleKey := range computePlan.AggregatetupleKeys {
+		aggregatetuple, err := getOutputAggregatetuple(db, aggregatetupleKey)
+		if err != nil {
+			return "", err
+		}
+
+		statusCollection = append(statusCollection, aggregatetuple.Status)
+	}
+
+	for _, traintupleKey := range computePlan.TraintupleKeys {
+		traintuple, err := getOutputTraintuple(db, traintupleKey)
+		if err != nil {
+			return "", err
+		}
+
+		statusCollection = append(statusCollection, traintuple.Status)
+	}
+
+	for _, compositeTraintupleKey := range computePlan.CompositeTraintupleKeys {
+		compositeTraintuple, err := getOutputCompositeTraintuple(db, compositeTraintupleKey)
+		if err != nil {
+			return "", err
+		}
+
+		statusCollection = append(statusCollection, compositeTraintuple.Status)
+	}
+
+	return determineComputePlanStatus(statusCollection)
+}
+
+func determineComputePlanStatus(statusCollection []string) (status string, err error) {
+	// this status order matters
+	sts := []string{StatusCanceled, StatusFailed, StatusDoing, StatusTodo, StatusWaiting, StatusDone}
+	for _, s := range sts {
+		if stringInSlice(s, statusCollection) {
+			return s, nil
+		}
+	}
+
+	return "", fmt.Errorf("unknown compute plan status")
+}
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -304,35 +304,24 @@ func getComputePlan(db *LedgerDB, key string) (resp outputComputePlan, err error
 		}
 	}
 
+	status, _ := getComputePlanStatus(db, aggregatetupleKeys, traintupleKeys, compositeTraintupleKeys)
+
 	resp = outputComputePlan{
 		ComputePlanID:           key,
 		TraintupleKeys:          traintupleKeys,
 		CompositeTraintupleKeys: compositeTraintupleKeys,
 		AggregatetupleKeys:      aggregatetupleKeys,
 		TesttupleKeys:           testtupleKeys,
+		Status:                  status,
 	}
 	return
 }
 
-func getComputePlanStatus(db *LedgerDB, args []string) (status string, err error) {
-	inp := struct {
-		ComputePlanID string `validate:"required,lte=64" json:"computePlanID"`
-	}{}
-
-	err = AssetFromJSON(args, &inp)
-	if err != nil {
-		return
-	}
-
-	computePlan, err := getComputePlan(db, inp.ComputePlanID)
-	if err != nil {
-		return "", err
-	}
-
+func getComputePlanStatus(db *LedgerDB, aggregatetupleKeys, traintupleKeys, compositeTraintupleKeys []string) (status string, err error) {
 	// get all tuples like status in one slice
 	statusCollection := []string{}
 
-	for _, aggregatetupleKey := range computePlan.AggregatetupleKeys {
+	for _, aggregatetupleKey := range aggregatetupleKeys {
 		aggregatetuple, err := getOutputAggregatetuple(db, aggregatetupleKey)
 		if err != nil {
 			return "", err
@@ -341,7 +330,7 @@ func getComputePlanStatus(db *LedgerDB, args []string) (status string, err error
 		statusCollection = append(statusCollection, aggregatetuple.Status)
 	}
 
-	for _, traintupleKey := range computePlan.TraintupleKeys {
+	for _, traintupleKey := range traintupleKeys {
 		traintuple, err := getOutputTraintuple(db, traintupleKey)
 		if err != nil {
 			return "", err
@@ -350,7 +339,7 @@ func getComputePlanStatus(db *LedgerDB, args []string) (status string, err error
 		statusCollection = append(statusCollection, traintuple.Status)
 	}
 
-	for _, compositeTraintupleKey := range computePlan.CompositeTraintupleKeys {
+	for _, compositeTraintupleKey := range compositeTraintupleKeys {
 		compositeTraintuple, err := getOutputCompositeTraintuple(db, compositeTraintupleKey)
 		if err != nil {
 			return "", err
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -309,7 +309,11 @@ func getComputePlan(db *LedgerDB, key string) (resp outputComputePlan, err error
 		}
 	}
 
-	status, err := getComputePlanStatus(db, tupleKeys)
+	keys := []string{}
+	keys = append(keys, tupleKeys...)
+	keys = append(keys, testtupleKeys...)
+
+	status, err := getComputePlanStatus(db, keys)
 	if err != nil {
 		return
 	}
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -304,8 +304,10 @@ func getComputePlan(db *LedgerDB, key string) (resp outputComputePlan, err error
 		}
 	}
 
-
-	status, _ := getComputePlanStatus(db, tupleKeys)
+	status, err := getComputePlanStatus(db, tupleKeys)
+	if err != nil {
+		return
+	}
 
 	resp = outputComputePlan{
 		ComputePlanID:           key,
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -364,29 +364,29 @@ func determineComputePlanStatus(statusCollection []string) (status string, err e
 	return "", fmt.Errorf("unknown compute plan status")
 }
 
-func cancelComputePlan(db *LedgerDB, args []string) (resp bool, err error) {
+func cancelComputePlan(db *LedgerDB, args []string) (resp outputComputePlan, err error) {
 	computeplan, err := queryComputePlan(db, args)
 	if err != nil {
-		return
+		return outputComputePlan{}, err
 	}
 
 	tupleKeys, err := db.GetIndexKeys("computePlan~computeplanid~worker~rank~key", []string{"computePlan", computeplan.ComputePlanID})
 	if err != nil {
-		return
+		return outputComputePlan{}, err
 	}
 	if len(tupleKeys) == 0 {
 		err = errors.E("no task found for compute plan %s", computeplan.ComputePlanID)
-		return
+		return outputComputePlan{}, err
 	}
 
 	for _, key := range tupleKeys {
 		tuple, err := db.GetGenericTuple(key)
 		if err != nil {
-			return false, err
+			return outputComputePlan{}, err
 		}
 		err = tuple.commitStatusUpdate(db, key, StatusCanceled)
 		if err != nil {
-			return false, err
+			return outputComputePlan{}, err
 		}
 	}
 
@@ -394,7 +394,7 @@ func cancelComputePlan(db *LedgerDB, args []string) (resp bool, err error) {
 	for _, key := range tupleKeys {
 		keys, err := db.GetIndexKeys("testtuple~traintuple~certified~key", []string{"testtuple", key})
 		if err != nil {
-			return false, err
+			return outputComputePlan{}, err
 		}
 
 		testtupleKeys = append(testtupleKeys, keys...)
@@ -403,13 +403,15 @@ func cancelComputePlan(db *LedgerDB, args []string) (resp bool, err error) {
 	for _, key := range testtupleKeys {
 		testtuple, err := db.GetTesttuple(key)
 		if err != nil {
-			return false, err
+			return outputComputePlan{}, err
 		}
 		err = testtuple.commitStatusUpdate(db, key, StatusCanceled)
 		if err != nil {
-			return false, err
+			return outputComputePlan{}, err
 		}
 	}
 
-	return true, nil
+	computeplan.Status = StatusCanceled
+
+	return computeplan, nil
 }
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -329,6 +329,21 @@ func getComputePlan(db *LedgerDB, key string) (resp outputComputePlan, err error
 	return
 }
 
+func getComputePlanStatusByComputePlanID(db *LedgerDB, computePlanID string) (status string, err error) {
+	computePlan, err := getComputePlan(db, computePlanID)
+	if err != nil {
+		return "", err
+	}
+
+	keys := []string{}
+	keys = append(keys, computePlan.TraintupleKeys...)
+	keys = append(keys, computePlan.CompositeTraintupleKeys...)
+	keys = append(keys, computePlan.AggregatetupleKeys...)
+	keys = append(keys, computePlan.TesttupleKeys...)
+
+	return getComputePlanStatus(db, keys)
+}
+
 func getComputePlanStatus(db *LedgerDB, keys []string) (status string, err error) {
 	// get all tuples like status in one slice
 	statusCollection := []string{}
@@ -394,11 +409,11 @@ func cancelComputePlan(db *LedgerDB, args []string) (resp bool, err error) {
 	}
 
 	for _, key := range testtupleKeys {
-		tuple, err := db.GetTesttuple(key)
+		testtuple, err := db.GetTesttuple(key)
 		if err != nil {
 			return false, err
 		}
-		err = tuple.commitStatusUpdate(db, key, StatusCanceled)
+		err = testtuple.commitStatusUpdate(db, key, StatusCanceled)
 		if err != nil {
 			return false, err
 		}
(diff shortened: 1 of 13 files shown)


chaincode/utils.go
```diff
@@ -155,3 +155,13 @@ func (modelType CompositeModelType) String() string {
...
+
+// Contains check if an element is contained in a slice of string
+func Contains(a []string, x string) bool {
+	for _, s := range a {
+		if s == x {
+			return true
+		}
+	}
+	return false
+}
```
thibaultrobert: There is already a function `stringInSlice` in `utils.go` that do exactly that.

chaincode/compute_plan.go
```diff
@@ -311,13 +311,62 @@ func getComputePlan(db *LedgerDB, key string) (resp outputComputePlan, err error
...
 	}
 	return
 }
+
+func getComputePlanStatus(db *LedgerDB, aggregatetupleKeys, traintupleKeys, compositeTraintupleKeys []string) (status string, err error) {
+	// get all tuples like status in one slice
+	statusCollection := []string{}
+
+	for _, aggregatetupleKey := range aggregatetupleKeys {
+		aggregatetuple, err := getOutputAggregatetuple(db, aggregatetupleKey)
```
thibaultrobert: Why getting the output tuples here? If there is no specific need you should only get the intern object (here it will be `db.GetAggregatetuple`). More over it exists a function `db.GetGenericTuple` that work for every tuple and allow you to acces common part of tuples including the status.

inelgnu: fixed here ef97caf

chaincode/compute_plan.go
```diff
@@ -311,13 +311,45 @@ func getComputePlan(db *LedgerDB, key string) (resp outputComputePlan, err error
 		}
 	}
 
+
+	status, _ := getComputePlanStatus(db, tupleKeys)
```
thibaultrobert: We should check the error here even for the status computation. Also I think it would be good to return the compute plan at creation.

AurelienGasser:
@inalgnu Sorry about the duplicate work ! (https://github.com/SubstraFoundation/substra-chaincode/pull/52)

Feel free to cherry pick some changes from there. I think the only valuable things to potentially bring over are 

- [these changes](https://github.com/SubstraFoundation/substra-chaincode/pull/52/files#diff-f0db4dc81b7e8c20cb3ecbea544ced0bR161-R293) to the composite/aggregate workflow test. 
- maybe also [these extra test cases](https://github.com/SubstraFoundation/substra-chaincode/pull/52/files#diff-f0db4dc81b7e8c20cb3ecbea544ced0bR500-R543)

Note that in https://github.com/SubstraFoundation/substra-chaincode/pull/52, and unlike this PR, we don't fetch _all_ the tuples from the ledger before determining the status of the compute plan. The _only_ advantage to doing that is for the "Canceled" case: we can fetch only one tuple and immediately return that the compute plan is "Canceled", without fetching any of the other tuples. I think this small optimization is not worth it, considering it makes the code (including tests) more complicated. Do you agree ? cc @thibaultrobert

thibaultrobert:
> Note that in #52, and unlike this PR, we don't fetch _all_ the tuples from the ledger before determining the status of the compute plan. The _only_ advantage to doing that is for the "Canceled" case: we can fetch only one tuple and immediately return that the compute plan is "Canceled", without fetching any of the other tuples. I think this small optimization is not worth it, considering it makes the code (including tests) more complicated. Do you agree ?

Like you, I tend to think it would be too costly a optimization even though it would reduce the read-set in the `queryComputePlan` smartcontract. And even if it was a bit complex to read, I did enjoy the function injection in your solution.

thibaultrobert:
I took upon myself to update the `README.md` to fix the tests @inalgnu. And I will also set the compute plan status in the `createComputePlan` smartcontract. Fill free to revert if you disagree with any of my commits.

thibaultrobert:
I have a question on the compute plan status:
The status of a compute plan is only compute (yes I just did that) from the train/composite/aggregate tuples and not from the testuples. So a compute plan can be `Done` even if one of the testtuples is `Doing` or even `Failed`.  Is that what was intended? (I didn't find the answer in the [spec](https://github.com/owkin/owkin-connect-spec/blob/master/motiv_modif/fail_fast.md), sorry if I missed it)

inelgnu:
> I have a question on the compute plan status:
> The status of a compute plan is only compute (yes I just did that) from the train/composite/aggregate tuples and not from the testuples. So a compute plan can be `Done` even if one of the testtuples is `Doing` or even `Failed`. Is that what was intended? (I didn't find the answer in the [spec](https://github.com/owkin/owkin-connect-spec/blob/master/motiv_modif/fail_fast.md), sorry if I missed it)

You're right, fixed here 94b0213

chaincode/testtuple.go
```diff
@@ -264,21 +266,35 @@ func logStartTest(db *LedgerDB, args []string) (outputTesttuple outputTesttuple,
...
+	tuple, err := db.GetGenericTuple(testtuple.TraintupleKey)
+	if err != nil {
+		return outputTesttuple{}, err
+	}
+
+	// cancel testtuple if compute plan is canceled
+	if tuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, tuple.ComputePlanID, &testtuple)
+		if err != nil {
+			return outputTesttuple{}, err
```
AurelienGasser, line 284: I think you can just do `return` without arguments here

chaincode/testtuple.go
```diff
@@ -264,21 +266,35 @@ func logStartTest(db *LedgerDB, args []string) (outputTesttuple outputTesttuple,
 	if err != nil {
 		return
 	}
+
+	tuple, err := db.GetGenericTuple(testtuple.TraintupleKey)
+	if err != nil {
+		return outputTesttuple{}, err
```
AurelienGasser, line 277: I think you can just do `return` without arguments here

chaincode/testtuple.go
```diff
@@ -290,6 +306,19 @@ func logSuccessTest(db *LedgerDB, args []string) (outputTesttuple outputTesttupl
 		return
 	}
 
+	tuple, err := db.GetGenericTuple(testtuple.TraintupleKey)
+	if err != nil {
+		return outputTesttuple{}, err
```
AurelienGasser, line 320: I think you can just do `return` without arguments here

chaincode/testtuple.go
```diff
@@ -290,6 +306,19 @@ func logSuccessTest(db *LedgerDB, args []string) (outputTesttuple outputTesttupl
...
+	}
+
+	// cancel testtuple if compute plan is canceled
+	if tuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, tuple.ComputePlanID, &testtuple)
+		if err != nil {
+			return outputTesttuple{}, err
```
AurelienGasser, line 327: I think you can just do `return` without arguments here

chaincode/testtuple.go
```diff
@@ -317,6 +346,19 @@ func logFailTest(db *LedgerDB, args []string) (outputTesttuple outputTesttuple,
 		return
 	}
 
+	tuple, err := db.GetGenericTuple(testtuple.TraintupleKey)
+	if err != nil {
+		return outputTesttuple{}, err
```
AurelienGasser, line 364: I think you can just do `return` without arguments here

chaincode/testtuple.go
```diff
@@ -317,6 +346,19 @@ func logFailTest(db *LedgerDB, args []string) (outputTesttuple outputTesttuple,
...
+	}
+
+	// cancel testtuple if compute plan is canceled
+	if tuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, tuple.ComputePlanID, &testtuple)
+		if err != nil {
+			return outputTesttuple{}, err
```
AurelienGasser, line 371: I think you can just do `return` without arguments here

chaincode/traintuple.go
```diff
@@ -280,19 +280,28 @@ func logStartTrain(db *LedgerDB, args []string) (outputTraintuple outputTraintup
 	if err != nil {
 		return
 	}
+
+	// cancel traintuple if compute plan is canceled
+	if traintuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, traintuple.ComputePlanID, &traintuple)
+		if err != nil {
+			return outputTraintuple{}, err
```
AurelienGasser, line 279: I think you can just do `return` without arguments here

chaincode/traintuple.go
```diff
@@ -305,6 +314,14 @@ func logSuccessTrain(db *LedgerDB, args []string) (outputTraintuple outputTraint
 	if err != nil {
 		return
 	}
+	// cancel traintuple if compute plan is canceled
+	if traintuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, traintuple.ComputePlanID, &traintuple)
+		if err != nil {
+			return outputTraintuple{}, err
```
AurelienGasser, line 316: I think you can just do `return` without arguments here

chaincode/traintuple.go
```diff
@@ -350,6 +367,14 @@ func logFailTrain(db *LedgerDB, args []string) (outputTraintuple outputTraintupl
...
```
AurelienGasser, line 372: I think you can just do `return` without arguments here

chaincode/traintuple_composite.go
```diff
@@ -301,105 +301,132 @@ func createCompositeTraintupleInternal(db *LedgerDB, inp inputCompositeTraintupl
...
 	if err != nil {
 		return
 	}
-	if err = validateTupleOwner(db, traintuple.Dataset.Worker); err != nil {
+
+	// cancel compositeTraintuple if compute plan is canceled
+	if compositeTraintuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, compositeTraintuple.ComputePlanID, &compositeTraintuple)
+		if err != nil {
+			return outputCompositeTraintuple{}, err
```
AurelienGasser, line 312: I think you can just do `return` without arguments here

chaincode/traintuple_composite.go
```diff
@@ -301,105 +301,132 @@ func createCompositeTraintupleInternal(db *LedgerDB, inp inputCompositeTraintupl
...
 		return
 	}
-	traintuple.Perf = inp.Perf
-	traintuple.OutHeadModel.OutModel = &HashDress{
+
+	// cancel compositeTraintuple if compute plan is canceled
+	if compositeTraintuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, compositeTraintuple.ComputePlanID, &compositeTraintuple)
+		if err != nil {
+			return outputCompositeTraintuple{}, err
```
AurelienGasser, line 350: I think you can just do `return` without arguments here

chaincode/traintuple_composite.go
```diff
@@ -301,105 +301,132 @@ func createCompositeTraintupleInternal(db *LedgerDB, inp inputCompositeTraintupl
...
 		return
 	}
-	traintuple.Log += inp.Log
 
-	if err = validateTupleOwner(db, traintuple.Dataset.Worker); err != nil {
+	// cancel compositeTraintuple if compute plan is canceled
+	if compositeTraintuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, compositeTraintuple.ComputePlanID, &compositeTraintuple)
+		if err != nil {
+			return outputCompositeTraintuple{}, err
```
AurelienGasser, line 406: I think you can just do `return` without arguments here

chaincode/tuple_aggregate.go
```diff
@@ -295,18 +295,27 @@ func logStartAggregate(db *LedgerDB, args []string) (outputAggregatetuple output
 	if err != nil {
 		return
 	}
+
+	// cancel aggregatetuple if compute plan is canceled
+	if aggregatetuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, aggregatetuple.ComputePlanID, &aggregatetuple)
+		if err != nil {
+			return outputAggregatetuple{}, err
```
AurelienGasser, line 285: I think you can just do `return` without arguments here

chaincode/tuple_aggregate.go
```diff
@@ -318,6 +327,15 @@ func logFailAggregate(db *LedgerDB, args []string) (outputAggregatetuple outputA
...
```
AurelienGasser, line 321: ok that's wayyy to many times I copy/pasted my comment already :D I'll stop now.

thibaultrobert: It would work at every time you copy/pasted it indeed but I think it's intentional here. @inalgnu will have to confirm it but return without explicit arguments are considered confusing and to be avoid.
Named return values are for documentation mostly but even then you should specified the return arguments. Here it would be `return o, err` at the very least.
It's mostly my bad if you've find a lot of bare return around the base code. Sorry about that.

AurelienGasser: Interesting, thanks for the precision! Let's see what @inalgnu says.

inelgnu: @AurelienGasser named params could be easy to use but might cause some issues especially shadowing. I used to avoid using them as a good practice.
this won't compile for example => https://play.golang.org/p/2rumK4hdKec
from codereviewcomments:
https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

> Don't name result parameters just to avoid declaring a var inside the function; that trades off a minor implementation brevity at the cost of unnecessary API verbosity.

In addition, here outputAggregatetuple type was named same as the parameter name which leads to an error if you want to create a new struct outputAggregatetuple{}
https://play.golang.org/p/g0RuGAR6W9t

Erratum:
Here is an example of shadowing, thank's to @thibaultrobert feedback on the first example which doesn't compile and it's a good news 😊 
https://play.golang.org/p/KL0mx9pvOo1

inelgnu:
@AurelienGasser  Sorry for the delay 😞 

> @inalgnu Sorry about the duplicate work ! (#52)
> 
> Feel free to cherry pick some changes from there. I think the only valuable things to potentially bring over are
> 
> * [these changes](https://github.com/SubstraFoundation/substra-chaincode/pull/52/files#diff-f0db4dc81b7e8c20cb3ecbea544ced0bR161-R293) to the composite/aggregate workflow test.
> * maybe also [these extra test cases](https://github.com/SubstraFoundation/substra-chaincode/pull/52/files#diff-f0db4dc81b7e8c20cb3ecbea544ced0bR500-R543)
> 
@AurelienGasser i would cherry-pick the tests with pleasure. Could you please separate them in atomic commits, or just commit them directly on this PR. Thank's :-) 

> Note that in #52, and unlike this PR, we don't fetch _all_ the tuples from the ledger before determining the status of the compute plan. The _only_ advantage to doing that is for the "Canceled" case: we can fetch only one tuple and immediately return that the compute plan is "Canceled", without fetching any of the other tuples. I think this small optimization is not worth it, considering it makes the code (including tests) more complicated. Do you agree ? cc @thibaultrobert

Your approach is way better, however, I'm not sure it worth it, for now, maybe we should measure before optimizing. What do you think @thibaultrobert ?

thibaultrobert:
> Your approach is way better, however, I'm not sure it worth it, for now, maybe we should measure before optimizing. What do you think @thibaultrobert ?

Like I answer before, I don't think this optimization is worth its load for now. We should measure first as you said.

chaincode/compute_plan.go
```diff
@@ -220,6 +224,7 @@ func createComputePlanInternal(db *LedgerDB, inp inputComputePlan) (resp outputC
 		resp.TesttupleKeys = append(resp.TesttupleKeys, testtupleKey)
```
thibaultrobert, line 217: We should add the testtuples' keys here I think.

chaincode/compute_plan.go
```diff
@@ -311,13 +316,116 @@ func getComputePlan(db *LedgerDB, key string) (resp outputComputePlan, err error
...
+		return "", err
+	}
+
+	keys := []string{}
+	keys = append(keys, computePlan.TraintupleKeys...)
+	keys = append(keys, computePlan.CompositeTraintupleKeys...)
+	keys = append(keys, computePlan.AggregatetupleKeys...)
+	keys = append(keys, computePlan.TesttupleKeys...)
+
+	return getComputePlanStatus(db, keys)
```
thibaultrobert: Now that I've read it like that I tend to think that `getComputePlan` should perhaps take the compute plan directly as argument and we should reconstruct this list in the function. What do you think @inalgnu

chaincode/compute_plan.go
thibaultrobert: After a quick exchange with @inalgnu , in the case of the `createComputePlan`, all the testtuples are supposed to be in a `waitingStatus` so there are not relevant to determine the compute plan global status. 
Sorry for the noise.

chaincode/compute_plan.go
inelgnu: @thibaultrobert that make sens, updated

chaincode/compute_plan.go
```diff
@@ -319,5 +320,104 @@ func getComputePlan(db *LedgerDB, key string) (resp outputComputePlan, err error
...
+		if err != nil {
+			return false, err
+		}
+		err = testtuple.commitStatusUpdate(db, key, StatusCanceled)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	return true, nil
```
thibaultrobert: Like I said directly to @inalgnu, the return value of a smartcontract should respect the [interface `Unmarshaler`](https://golang.org/pkg/encoding/json/#Unmarshaler).

chaincode/ledger.go
```diff
@@ -37,6 +37,11 @@ const (
 	// the String() function in utils.go
 )
 
+// StatusUpdater is exported
+type StatusUpdater interface {
```
thibaultrobert, line 41: The power of interface has come among us. 👍

chaincode/testtuple.go
```diff
@@ -159,6 +159,8 @@ func (testtuple *Testtuple) SetFromTraintuple(db *LedgerDB, traintupleKey string
 		return errors.BadRequest(
 			"could not register this testtuple, the traintuple %s has a failed status",
 			traintupleKey)
```
thibaultrobert, line 165: That has nothing to do with this PR per say but we are updating the different train like tuples to be created in a `StatusFailed` if their inModels are failed. Do we want a similar approach for testtuples? (Here we return an error, we could create the testtuple in a failed state)

chaincode/testtuple.go
```diff
@@ -264,21 +266,35 @@ func logStartTest(db *LedgerDB, args []string) (outputTesttuple outputTesttuple,
...
 	}
+
+	tuple, err := db.GetGenericTuple(testtuple.TraintupleKey)
+	if err != nil {
+		return outputTesttuple{}, err
+	}
+
+	// cancel testtuple if compute plan is canceled
+	if tuple.ComputePlanID != "" {
+		err := cancelIfComputePlanIsCanceled(db, inp.Key, tuple.ComputePlanID, &testtuple)
```
thibaultrobert: I think there is a problem here: we should return after canceling this tuple. If we continue, we will update the status later in the function to `StatusDoing` or am I missing something obvious?
And this apply to every other `LogXXX` function.

chaincode/compute_plan.go
inelgnu: done, thank's

chaincode/tuple_aggregate.go
AurelienGasser: Thank you, and sorry for the avalanche of comments!

chaincode/testtuple.go
inelgnu: yes, I will fix that

inelgnu: fixed here 681506d

thibaultrobert:
I've just made a attempt to keep the interface approach to update the status. The problem with the current approach is that I created a getter on an interface and I'm not 100% comfortable with that.
What do you think?
A partial Update of the JSON is also a possible solution to update status with just `GenericTuple`

thibaultrobert (approved):

chaincode/ledger.go
```diff
@@ -133,6 +138,7 @@ type CompositeTraintuple struct {
 	InTrunkModel  string                      `json:"inTrunkModel"`
 	OutHeadModel  CompositeTraintupleOutModel `json:"outHeadModel"`
 	OutTrunkModel CompositeTraintupleOutModel `json:"outTrunkModel"`
+	Perf          float32                     `json:"perf"`
```
AurelienGasser, line 141: @inalgnu I just noticed this line, is this addition intentional?

inelgnu: nope this seems to be due to a bad rebase :-/

AurelienGasser: https://github.com/SubstraFoundation/substra-chaincode/pull/61

PR #58, opened 2019-12-12
Move examples to their own file

AurelienGasser:
- move the generation of the examples to a new file `generate_examples_test.go` (instead of being a part of `main_test.go`
- change the default examples file to be `EXAMPLES.md` instead of `README.md` (I'm not convinced myself by the name `EXAMPLES.md` so feel free to suggest something else!


thibaultrobert (dismissed):
Yes, yes and yes

thibaultrobert (approved):

PR #63, opened 2019-12-16
allow doing tuple update after compute plan cancel

inelgnu:


thibaultrobert (dismissed):
LGTM

chaincode/compute_plan.go
```diff
@@ -370,6 +370,10 @@ func cancelComputePlan(db *LedgerDB, args []string) (resp outputComputePlan, err
 		return outputComputePlan{}, err
 	}
 
+	if computeplan.Status == StatusCanceled {
```
thibaultrobert: you should also return early if `computeplan.Status` is either `StatusDone` or `StatusFailed` because it means that no tuples can be updated.
Moreover the return status value will differ between this smartcontract and a `queryComputePlan` one.

thibaultrobert (approved):

chaincode/compute_plan.go
inelgnu: done

PR #64, opened 2019-12-16
Add an event in case of compute plan

thibaultrobert:
Companion PR: https://github.com/SubstraFoundation/substra-backend/pull/90

### DIFF
--- a/chaincode/ledger_db.go
+++ b/chaincode/ledger_db.go
@@ -30,7 +30,7 @@ type State struct {
 // LedgerDB to access the chaincode database during the lifetime of a SmartContract
 type LedgerDB struct {
 	cc               shim.ChaincodeStubInterface
-	tuplesEvent      *TuplesEvent
+	event            *Event
 	transactionState State
 	mutex            *sync.RWMutex
 }
@@ -443,14 +443,14 @@ func (db *LedgerDB) GetNode(key string) (Node, error) {
 // SendTuplesEvent sends an event with updated tuples if there is any
 // Only one event can be sent per transaction
 func (db *LedgerDB) SendTuplesEvent() error {
-	if db.tuplesEvent == nil {
+	if db.event == nil {
 		return nil
 	}
-	payload, err := json.Marshal(*(db.tuplesEvent))
+	payload, err := json.Marshal(*(db.event))
 	if err != nil {
 		return err
 	}
-	err = db.cc.SetEvent("tuples-updated", payload)
+	err = db.cc.SetEvent("chaincode-updates", payload)
 	if err != nil {
 		return err
 	}
@@ -469,8 +469,8 @@ func (db *LedgerDB) AddTupleEvent(tupleKey string) error {
 	if genericTuple.Status != StatusTodo {
 		return nil
 	}
-	if db.tuplesEvent == nil {
-		db.tuplesEvent = &TuplesEvent{}
+	if db.event == nil {
+		db.event = &Event{}
 	}
 	switch genericTuple.AssetType {
 	case TraintupleType:
@@ -480,7 +480,7 @@ func (db *LedgerDB) AddTupleEvent(tupleKey string) error {
 		}
 		out := outputTraintuple{}
 		out.Fill(db, tuple, tupleKey)
-		db.tuplesEvent.Traintuples = append(db.tuplesEvent.Traintuples, out)
+		db.event.Traintuples = append(db.event.Traintuples, out)
 	case CompositeTraintupleType:
 		tuple, err := db.GetCompositeTraintuple(tupleKey)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


thibaultrobert:
@samlesu @AurelienGasser @inalgnu 
I've just added a basic compute plan event in case of success. I've done it in a very simple fashion so it didn't change the behavior of other element of the event.
Do you think we should keep this structure for now or do you have an alternative to suggest?

samlesu:
> @samlesu @AurelienGasser @inalgnu
> I've just added a basic compute plan event in case of success. I've done it in a very simple fashion so it didn't change the behavior of other element of the event.
> Do you think we should keep this structure for now or do you have an alternative to suggest?

Thanks Thibault. Looks good to me.

chaincode/ledger_db.go
```diff
@@ -494,3 +494,13 @@ func (db *LedgerDB) AddTupleEvent(tupleKey string) error {
 	}
 	return nil
 }
+
+// AddComputePlanEvent add the output tuple matching the tupleKey to the event struct
```
AurelienGasser: I think you forgot to update the comment for the compute plan case

chaincode/output.go
```diff
@@ -279,12 +279,13 @@ type outputModel struct {
...
 
-// TuplesEvent is the collection of tuples sent in an event
-type TuplesEvent struct {
+// Events is the collection of tuples sent in an event
+type Events struct {
 	Testtuples           []outputTesttuple           `json:"testtuple"`
 	Traintuples          []outputTraintuple          `json:"traintuple"`
 	CompositeTraintuples []outputCompositeTraintuple `json:"compositeTraintuple"`
 	Aggregatetuples      []outputAggregatetuple      `json:"aggregatetuple"`
+	ComputePlan          []outputComputePlan         `json:"computePlan"`
```
AurelienGasser: Nit: The new property doesn't have an "s" but the existing properties do.

chaincode/ledger_db.go
AurelienGasser: I think you forgot to update the comment for the compute plan case

chaincode/output.go
AurelienGasser: Nit: The new property doesn't have an "s" but the existing properties do.

chaincode/output.go
```diff
@@ -279,12 +279,13 @@ type outputModel struct {
 	Testtuple           outputTesttuple            `json:"testtuple"`
 }
 
-// TuplesEvent is the collection of tuples sent in an event
-type TuplesEvent struct {
+// Events is the collection of tuples sent in an event
+type Events struct {
```
AurelienGasser: Is there a particular reason to add an s to the struct name? For me, having a "single" event with multiple things inside is easier to reason with. But maybe that's just me!

chaincode/tuple_aggregate.go
```diff
@@ -390,6 +390,18 @@ func logSuccessAggregate(db *LedgerDB, args []string) (o outputAggregatetuple, e
 		return
 	}
 
+	// send an compute plan event if this one is done
```
AurelienGasser: 1. Is there a reason why we don't send an event in the "failed" case?
2. If two tuples from the same CP get marked as failed, we're going to generate two events. Is that intentional?

AurelienGasser (changes requested):

chaincode/output.go
thibaultrobert: I add an internal struggle here. I agree that there is in fact only one event from hyperledger point of vue and only one structure but each item in this struct's slices is by itself an event.
But I would be ok with `Event`. @inalgnu @samlesu any strong feeling about that?

chaincode/tuple_aggregate.go
thibaultrobert: For now we only send an event if there is a success for computeplan. It's purpose is to remove the `/local` directory in the different workers. In other cases (`Failed` and `Cancelled`), it should not be removed automatically.

chaincode/ledger_db.go
thibaultrobert: done

chaincode/output.go
thibaultrobert: Done

thibaultrobert:
@AurelienGasser @inalgnu are you ok with this PR? Or do you think we should send an event in the other case (`Failed` and `Cancelled`).

AurelienGasser:
> are you ok with this PR? Or do you think we should send an event in the other case (`Failed` and `Cancelled`).

I have no strong opinion. The downside of _not_ generating an event is that the local folder will not be deleted if the compute plan fails or gets canceled.

inelgnu:
@thibaultrobert @AurelienGasser Let's send an event in all cases and let the backend decide which one should be handled.

thibaultrobert:
I close this old one because there has been too many changes to the compute plans. Let's start fresh.

Issue #65, opened 2019-12-16
[BUG] error while getting associated traintuples to update their inModel


jmorel:
**Describe the bug**

While creating a fairly large compute plan, my first tuple of the plan is stuck in doing while all others are stuck in waiting.

**Logs**  

Here are the chaincode logs after the creation of the compute plan.
```
2019-12-16 15:55:09.926 UTC [substra-chaincode] Infof -> INFO 02f Args received by the chaincode: []string{"logStartCompositeTrain", "{\"key\": \"a43a2efc1415b16618d1a1349eb9c89b5d6bee0849a3395bc9af3504cebad19d\"}"}
2019-12-16 15:55:09.931 UTC [substra-chaincode] Infof -> INFO 030 traintuple a43a2efc1415b16618d1a1349eb9c89b5d6bee0849a3395bc9af3504cebad19d status updated: doing (from=todo)
2019-12-16 15:55:09.931 UTC [substra-chaincode] Infof -> INFO 031 Response from chaincode: main.outputCompositeTraintuple{Key:"a43a2efc1415b16618d1a1349eb9c89b5d6bee0849a3395bc9af3504cebad19d", Algo:(*main.HashDressName)(0xc0001b8300), Creator:"MyOrg1MSP", Dataset:(*main.outputTtDataset)(0xc00033b140), ComputePlanID:"a43a2efc1415b16618d1a1349eb9c89b5d6bee0849a3395bc9af3504cebad19d", InHeadModel:(*main.Model)(nil), InTrunkModel:(*main.Model)(nil), Log:"", OutHeadModel:main.outModelComposite{OutModel:(*main.HashDress)(nil), Permissions:main.outputPermissions{Process:main.Permission{Public:false, AuthorizedIDs:[]string{"MyOrg1MSP"}}}}, OutTrunkModel:main.outModelComposite{OutModel:(*main.HashDress)(nil), Permissions:main.outputPermissions{Process:main.Permission{Public:false, AuthorizedIDs:[]string{"MyOrg1MSP"}}}}, Rank:0, Status:"doing", Tag:""}, error: %!s(<nil>)
[...]

PR #69, opened 2020-01-06
fix issue #65

inelgnu:


samlesu:
Well done the investigation and the work @thibaultrobert and @inalgnu. 

As far as I understand, these changes are solving two issues:
1. a recursion issue when creating tuple (bug #65)
2. timeout error on asset creation (linked to https://github.com/SubstraFoundation/substra-backend/issues/10)

Do you think it would be worth splitting this PR into two separate PRs? It could then be easier to track the changes and could also speed up the review/merge process faster.

samlesu:
I'm wondering the side effect of removing the conflict error when  creating a compute plan / traintuple.

- Would it be possible to create a compute plan depending on existing traintuples (= on existing compute plans)?
- If yes, we may have to handle a few corner cases:
  - compute plan depending on other compute plan: cleanup in the backend?
  - handle 'failed', 'canceled' statuses

inelgnu:
@samlesu as discussed I've kept conflict error on tuple creation.

samlesu (approved):
Thanks!

thibaultrobert (approved):

Issue #70, opened 2020-01-07
[BUG] Scale limit on compute plans


thibaultrobert:
**Describe the bug**
While working on debugging and solutioning the issue #65, we tried to create and cancel a compute plan with 3000 tuples on 2 nodes. It was a alternance of 2 composite traintuples ,one on each node, and one aggregate tuple on 1 node, the same every time. It correspond to [this test](https://github.com/SubstraFoundation/substra-tests/blob/59731e669d057ce12050abda079c57c9897ea4f3/tests/test_load.py#L401) with a `COMPUTE_PLAN_SIZES` set at 1000.

While doing so we encounter several problems due to the scale of such compute plan:

1. we can't list all the composite traintuples
2. we can't cancel the compute plan

In both case, a GRPC error complaining about a too big message such as:
```
Error: Request failed: InvalidRequest: 400 Client Error: Bad Request for url: http://substra-backend.node-1.com/composite_traintuple/: <_MultiThreadedRendezvous of RPC that terminated with:
        status = StatusCode.RESOURCE_EXHAUSTED
        details = "Received message larger than max (7775148 vs. 4194304)"
        debug_error_string = "{"created":"@1578402814.533204900","description":"Received message larger than max (7775148 vs. 4194304)","file":"src/core/ext/filters/message_size/message_size_filter.cc","file_line":191,"grpc_status":8}">
```

So both call obviously send through a non streaming GRPC a message exceeding the 4.1M limit set. But it was for different reason.

_Case 1: Too big of an object_

When we list the all the composite traintuples, we make a `Query` via the substra-backend to the hyperledger peer of one node.
This peer submit such `Query` to the chaincode which treat it and return the exhaustive list all at once. There is no writing in the ledger here, no endorsement or any other step involving an other node.

And the problem we're getting here is that the object we return is so massive, it makes the message send over GRPC exceed the size limit.

So we should rethink the way we handle this listing type of queries, perhaps with a pagination system.

_Case 2: Too big of a read/write set_

Here we tried to cancel a compute plan so we make an `Invoke` via the substra-backend to the hyperledger peer of one node.
This `Invoke` is then pass to the chaincode which will retrieve all the tuples of this compute plan and will update all of their status. Then it will return a visualization of this compute plan (the list of all its tuples' keys, its status and its ID).

[...]

PR #71, opened 2020-01-08
Transform the compute plan to a ledger asset

thibaultrobert:
To answer several problems encountered lately (#70 for instance), a decision to create an asset to represent a compute plan have been made.

Here's a none exhaustive list of change to do to make in works:

- [x] Chose the data struct that will be saved on the ledger:
    - lists of tuples keys (with or without rank information)
    - status
    - ...
- [x] Change the way a compute plan ID is created
    We can't use the first tuple key as ID since it's already an asset key. So we should either generate it randomly or use a salted version of the first tuple key if we want to keep a similar behavior.
- [x] Change the way the different tuples' status is determined and updated


chaincode/ledger.go
```diff
@@ -179,6 +180,15 @@ type Testtuple struct {
 	Tag           string      `json:"tag"`
 }
 
+type ComputePlan struct {
+	AggregatetupleKeys      []string  `json:"aggregatetupleKeys"`
+	AssetType               AssetType `json:"assetType"`
+	CompositeTraintupleKeys []string  `json:"compositeTraintupleKeys"`
+	Status                  string    `json:"status"`
+	TesttupleKeys           []string  `json:"testtupleKeys"`
+	TraintupleKeys          []string  `json:"traintupleKeys"`
```
inelgnu, line 192: I think we will need a `tags` or `labels` field for a compute_plan

samlesu: I also think it would be great to support multiple tags/labels.
We could do an another PR doing the change for all the existing assets having a `tag` field?

thibaultrobert: I agree it should be `tags` and not `tag` but I also agree it should be on a different PR. Even more so since it will break compatibility with the backend.

And for the `tags` property on compute plan, I will add it after successfully reproduce the current compute plan capability. Good idea !

Thanks for the feedbacks

chaincode/traintuple_composite.go
```diff
@@ -169,30 +169,36 @@ func (traintuple *CompositeTraintuple) AddToComputePlan(db *LedgerDB, inp inputC
 			err = errors.BadRequest("invalid inputs, a new ComputePlan should have a rank 0")
 			return err
 		}
-		traintuple.ComputePlanID = traintupleKey
+		// TODO: Creage a function to addle the status of a compute plan
```
inelgnu: the cat walked around the keyboard? 🐈

chaincode/tuple_aggregate.go
```diff
@@ -139,23 +139,30 @@ func (tuple *Aggregatetuple) AddToComputePlan(db *LedgerDB, inp inputAggregatetu
 			err = errors.BadRequest("invalid inputs, a new ComputePlan should have a rank 0")
 			return err
 		}
-		tuple.ComputePlanID = traintupleKey
+		// TODO: Creage a function to addle the status of a compute plan
```
inelgnu: 🐈

thibaultrobert:
Still need a little clean-up but it's working.

chaincode/compute_plan.go
```diff
@@ -109,10 +108,9 @@ func createComputePlan(db *LedgerDB, args []string) (resp outputComputePlan, err
 }
 
 func createComputePlanInternal(db *LedgerDB, inp inputComputePlan) (resp outputComputePlan, err error) {
```
thibaultrobert, line 126: I should re-work how this function interact with the newly created asset

thibaultrobert: Done

chaincode/tuple.go
```diff
@@ -30,6 +30,7 @@ const (
 	StatusFailed   = "failed"
 	StatusDone     = "done"
 	StatusCanceled = "canceled"
+	StatusAborted  = "canceled"
```
inelgnu: maybe add a comment here to explain why is it equal to canceled, for now, that could be helpful...

inelgnu (dismissed):
LGTM 👍

inelgnu (approved):

PR #74, opened 2020-01-20
Add counters to outputComputePlan

thibaultrobert:


inelgnu (dismissed):
LGTM, I think it would better to have some consistency between input and output on 'TupleCount' 'TuplesCount' field, but it's minor.

inelgnu (approved):

PR #78, opened 2020-01-22
Correct errors

### LINKED ISSUE #77: [BUG] Chaincode returns 500 error instead of 400 in some cases
**Describe the bug**
The chaincode responds in some cases a 500 error instead of a 400 error. For traceability and debugging it should be fixed: 500 errors are server errors and 400 errors are user errors.
 
**Investigation**

This is at least true for the following error:
```go
            return Permissions{}, fmt.Errorf("invalid permission input values")
```

We should go through all the errors returned by the chaincode and ensured they have the proper status code.

thibaultrobert:
Fixes #77

### DIFF
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -16,7 +16,6 @@ package main
 
 import (
 	"chaincode/errors"
-	"fmt"
 	"strconv"
 )
 
@@ -31,7 +30,7 @@ func (inpTraintuple *inputTraintuple) Fill(inpCP inputComputePlanTraintuple, tra
 	for _, InModelID := range inpCP.InModelsIDs {
 		inModelKey, ok := traintupleKeysByID[InModelID]
 		if !ok {
-			return fmt.Errorf("model ID %s not found", InModelID)
+			return errors.BadRequest("model ID %s not found", InModelID)
 		}
 		inpTraintuple.InModels = append(inpTraintuple.InModels, inModelKey)
 	}
@@ -49,7 +48,7 @@ func (inpAggregatetuple *inputAggregatetuple) Fill(inpCP inputComputePlanAggrega
 	for _, InModelID := range inpCP.InModelsIDs {
 		inModelKey, ok := aggregatetupleKeysByID[InModelID]
 		if !ok {
-			return fmt.Errorf("model ID %s not found", InModelID)
+			return errors.BadRequest("model ID %s not found", InModelID)
 		}
 		inpAggregatetuple.InModels = append(inpAggregatetuple.InModels, inModelKey)
 	}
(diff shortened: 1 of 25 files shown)


chaincode/compute_plan.go
```diff
@@ -31,7 +30,7 @@ func (inpTraintuple *inputTraintuple) Fill(inpCP inputComputePlanTraintuple, tra
 	for _, InModelID := range inpCP.InModelsIDs {
 		inModelKey, ok := traintupleKeysByID[InModelID]
 		if !ok {
-			return fmt.Errorf("model ID %s not found", InModelID)
+			return errors.BadRequest("model ID %s not found", InModelID)
```
inelgnu, line 33: shouldn't this be a NotFound error?

chaincode/compute_plan_dag.go
```diff
@@ -75,7 +75,7 @@ func (dag *ComputeDAG) sort() error {
 			current[i].Depth = depth
 			final = append(final, current[i])
 			if _, ok := IDPresents[current[i].ID]; ok {
-				return fmt.Errorf("compute plan error: Duplicate training task ID: %s", current[i].ID)
+				return errors.Internal("compute plan error: Duplicate training task ID: %s", current[i].ID)
```
inelgnu: This error seems to be a conflict, no?

chaincode/compute_plan.go
thibaultrobert: The ID are found in the input, so if it's missing it's due to the arguments so a BadRequest was more logical what do you think?

inelgnu: indeed

chaincode/compute_plan_dag.go
thibaultrobert: Yes it's a conflict problem... Will update it

thibaultrobert: Done

chaincode/compute_plan_dag.go
```diff
@@ -75,7 +75,7 @@ func (dag *ComputeDAG) sort() error {
 			current[i].Depth = depth
 			final = append(final, current[i])
 			if _, ok := IDPresents[current[i].ID]; ok {
-				return fmt.Errorf("compute plan error: Duplicate training task ID: %s", current[i].ID)
+				return errors.Conflict("compute plan error: Duplicate training task ID: %s", current[i].ID)
```
samlesu: This is a BadRequest? The request is invalid due to duplicate ids. What do you think?

chaincode/compute_plan_dag.go
```diff
@@ -90,7 +90,7 @@ func (dag *ComputeDAG) sort() error {
 			for _, c := range current {
 				errorIDs = append(errorIDs, c.ID)
 			}
-			return fmt.Errorf("compute plan error: Cyclic or missing dependency among inModels IDs: %v", errorIDs)
+			return errors.Internal("compute plan error: Cyclic or missing dependency among inModels IDs: %v", errorIDs)
```
samlesu: Could be a BadRequest as well: we don't support cyclic graph.

chaincode/compute_plan_dag.go
thibaultrobert: I would say it's debatable here because its still a conflict but I think BadRequest is more coherent with the rest so let's change it.

thibaultrobert: Done

chaincode/compute_plan_dag.go
thibaultrobert: Done

samlesu (approved):
LGTM

inelgnu (approved):

PR #79, opened 2020-01-23
Query model permissions

thibaultrobert:
Works with https://github.com/SubstraFoundation/substra-backend/pull/117


jmorel:
I tested this PR against https://github.com/SubstraFoundation/substra-backend/pull/117/files and it works like a charm!

README.md
```diff
@@ -34,6 +34,7 @@ Note for internal use only: See the [technical specifications](https://github.co
 
 ### Implemented smart contracts
 
+- `cancelComputePlan`
```
inelgnu, line 37: is this related to this PR?

thibaultrobert: Yes and no: its because I updated this list with the missing smart contract.

chaincode/tuple.go
```diff
@@ -256,3 +294,11 @@ func determineTupleStatus(db *LedgerDB, tupleStatus, computePlanID string) (stri
...
 	return tupleStatus, nil
 }
+
+func createModelIndex(db *LedgerDB, tupleKey, modelHash string) error {
+	err := db.CreateIndex("tuple~modelHash~key", []string{"tuple", modelHash, tupleKey})
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
AurelienGasser: unused

chaincode/tuple.go
```diff
@@ -163,6 +163,44 @@ func queryModels(db *LedgerDB, args []string) (outModels []outputModel, err erro
 	return
 }
 
+func queryModelPermissions(db *LedgerDB, args []string) (outputPermissions, error) {
```
AurelienGasser, line 166: It took me a long time to understand this function. I think more explicit naming would help make it more clear (see comments below)

chaincode/tuple.go
```diff
@@ -163,6 +163,44 @@ func queryModels(db *LedgerDB, args []string) (outModels []outputModel, err erro
...
+	var out outputPermissions
+	inp := inputKey{}
+	err := AssetFromJSON(args, &inp)
+	if err != nil {
+		return out, err
+	}
```
AurelienGasser, line 172: Here for instance, add:

```go
modelHash := inp.Key
```

and use `modelHash` throughout the function instead of `inp.Key`

chaincode/tuple.go
```diff
@@ -163,6 +163,44 @@ func queryModels(db *LedgerDB, args []string) (outModels []outputModel, err erro
...
+	keys, err := db.GetIndexKeys("tuple~modelHash~key", []string{"tuple", inp.Key})
+	if err != nil {
+		return out, err
+	}
+	if len(keys) == 0 {
+		return out, errors.NotFound("Could not find a model for hash %s", inp.Key)
+	}
+	key := keys[0]
```
AurelienGasser: Here too:

```suggestion
	tupleKey := keys[0]
```

chaincode/tuple.go
```diff
@@ -163,6 +163,44 @@ func queryModels(db *LedgerDB, args []string) (outModels []outputModel, err erro
...
+	modelPermissions := Permissions{}
+	modelPermissions.Process.Public = true
+
+	// get out-model and permissions from parent for head model in composite traintuple
+	if tupleType == CompositeTraintupleType{
+		tuple, err := db.GetCompositeTraintuple(key)
+		if err != nil {
+			return out, errors.Internal(err, "queryModelPermissions:")
+		}
+		if tuple.OutHeadModel.OutModel.Hash == inp.Key {
```
AurelienGasser: Maybe add a comment here? Something like 

```go
// if `modelHash` refers to the head out-model, return the head out-model permissions
// if `modelHash` refers to the trunk out-model, do nothing (default to "public processable")
```

chaincode/tuple_test.go
```diff
@@ -182,3 +182,17 @@ func TestTagTuple(t *testing.T) {
...
+	mockStub := NewMockStubWithRegisterNode("substra", scc)
+	registerItem(t, *mockStub, "traintuple")
+	mockStub.MockTransactionStart("42")
+	db := NewLedgerDB(mockStub)
+	traintupleToDone(t, db, traintupleKey)
+	outTrain, err := queryTraintuple(db, keyToArgs(traintupleKey))
+	assert.NoError(t, err)
+	outPerm, err := queryModelPermissions(db, keyToArgs(outTrain.OutModel.Hash))
+	assert.NoError(t, err)
+	assert.NotZero(t, outPerm)
```
AurelienGasser, line 197: IMO it would be worth adding a test that retrieves/validates permissions for a head out-model here, since (if my understanding is correct) the `queryModelPermissions` function's primary purpose is to make sure we don't try to process head out-models that don't belong to our worker.

AurelienGasser (changes requested):
Great work! There are a few things worth improving IMO (nothing big), see comments below.

chaincode/tuple.go
thibaultrobert: Yes indeed. To you think I should removed it or used it.
On one side it will not change the number of line but on the bright side this index name will not be copied for each creation (so 4 time in the actual base code)

chaincode/tuple.go
thibaultrobert: Further more I think we should use an `InputHash` here not an `inputKey` since it's indeed a hash. @Kelvin-M do you think we could update the backend in that direction?
I'll add `modelHash` for now.

chaincode/tuple.go
thibaultrobert: Yep and done.

chaincode/tuple.go
thibaultrobert: Added

chaincode/tuple_test.go
thibaultrobert: For now yes and it should be extended to all models in near future. I will update the test.

chaincode/tuple.go
Kelvin-M: Yes we can, but the modification must be done also for all object to  have the same usage. So let's do it later because it will be too long to verify that everything works :)

thibaultrobert:
Sorry for the noise, I push the wrong commit to the wrong branch.

chaincode/tuple.go
thibaultrobert: Done

chaincode/tuple.go
thibaultrobert: Done

chaincode/tuple.go
thibaultrobert: I was speaking about this smart contract in particular since `queryModelPermissions` returns the Permission of the Model matching the `Hash` in the arguments.

chaincode/tuple_test.go
thibaultrobert: Done in f3f4c7b53ea1114ad551505c2add5724a2dbfe7b.

AurelienGasser (approved):

PR #81, opened 2020-01-28
Asset add: return structure instead of key

samlesu:


chaincode/traintuple.go
```diff
@@ -210,20 +210,24 @@ func (traintuple *Traintuple) Save(db *LedgerDB, traintupleKey string) error {
...
+		return
 	}
 
-	return map[string]string{"key": key}, nil
+	traintuple, err := db.GetTraintuple(key)
+	if err != nil {
+		return
+	}
+	err = out.Fill(db, traintuple, key)
+	return
```
thibaultrobert: I think you should change the `createTraintupleInternal` so it return the traintuple it just saved. It would prevent one more de-serialization.

chaincode/traintuple.go
```diff
@@ -210,20 +210,24 @@ func (traintuple *Traintuple) Save(db *LedgerDB, traintupleKey string) error {
 // -------------------------------------------------------------------------------------------
 
 // createTraintuple adds a Traintuple in the ledger
-func createTraintuple(db *LedgerDB, args []string) (map[string]string, error) {
+func createTraintuple(db *LedgerDB, args []string) (out outputTraintuple, err error) {
```
thibaultrobert: Why do you go back to named return values?

samlesu: No reason, I will restore the unnamed return values.

chaincode/traintuple.go
samlesu: Not 100% sure as `createTraintupleInternal` is also used inside a compute plan.

thibaultrobert: Sadly for now you right, forget I said anything. 
But in this case I think the `err` should be wrap in an Internal error since its clearly not normal it it doesn't find it. What do you think?

Kelvin-M:
@samlesu Is it still relevant ?

samlesu:
Nop, it's too old now :) Thanks

PR #87, opened 2020-02-03
Prevent from composite traintuple creation if worker not the same

Kelvin-M:


jmorel (approved):
Tested via https://github.com/SubstraFoundation/substra-tests/pull/74 and working as expected!

PR #92, opened 2020-02-12
Gzip ledger assets

AurelienGasser:
Store the ledger assets as Gzipped JSON (instead of just JSON)

Based off benchmark:

```
json       | 788 ms   | 1208 bytes
json-gzip  | 12112 ms | 405 bytes
gob        | 3956 ms  | 1391 bytes
gob-gzip   | 17545 ms | 602 bytes
```
- tested on a `CompositeTraintuple`
- time column: serialize/deserialize 50k times
- weight column: weight of a serialized `CompositeTraintuple`
- benchmark source : https://github.com/SubstraFoundation/substra-chaincode/commit/e37e3a7e078d18d8bc1bf7df5d552770d6226233

### DIFF
--- a/chaincode/ledger_db.go
+++ b/chaincode/ledger_db.go
@@ -15,15 +15,20 @@
 package main
 
 import (
-	"bytes"
 	"chaincode/errors"
-	"compress/gzip"
 	"encoding/json"
-	"io/ioutil"
 	"sync"
 
 	"github.com/hyperledger/fabric/core/chaincode/shim"
 )
+
+var ledgerSettings LedgerSettings
+
+// LedgerSettings ...
+type LedgerSettings struct {
+	EnableGzip bool   `json:"enableGzip"`
+	Foo        string `json:"foo"`
+}
 
 // State is a in-memory representation of the db state
 type State struct {
@@ -75,19 +80,26 @@ func (db *LedgerDB) putTransactionState(key string, state []byte) {
 
 // Get retrieves an object stored in the chaincode db and set the input object value
 func (db *LedgerDB) Get(key string, object interface{}) error {
-	var buff []byte
-	var err error
+	var jsonBytes []byte
 
-	buff, ok := db.getTransactionState(key)
+	jsonBytes, ok := db.getTransactionState(key)
 	if !ok {
-		buff, err = db.GetStateGzip(key)
-		if err != nil || buff == nil {
+		ledgerBytes, err := db.cc.GetState(key)
+		if err != nil || ledgerBytes == nil {
 			return errors.NotFound(err, "no asset for key %s", key)
 		}
-		db.putTransactionState(key, buff)
+		if ledgerSettings.EnableGzip {
+			jsonBytes, err = UnGzip(ledgerBytes)
+			if err != nil {
+				return errors.Internal(err, "error while ungzipping data for key %s", key)
+			}
+		} else {
+			jsonBytes = ledgerBytes
+		}
+		db.putTransactionState(key, jsonBytes)
 	}
 
-	return json.Unmarshal(buff, &object)
+	return json.Unmarshal(jsonBytes, &object)
 }
 
 // KeyExists checks if a key is stored in the chaincode db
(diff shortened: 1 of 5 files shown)


AurelienGasser:
I haven't tested with `make test` because tests currently don't pass on `master` on my machine :'(

thibaultrobert (dismissed):

inelgnu:
@AurelienGasser @thibaultrobert keep in mind that this won't work if we switch the storage to CouchDB. So we should maybe add a reminder comment or make this feature optional.

AurelienGasser:
@inalgnu Good point! I looked into it and I [found this](https://hyperledger-fabric.readthedocs.io/en/release-1.4/couchdb_as_state_database.html):

> If you have a mix of JSON and binary data values, you can still use CouchDB, however the binary values can only be queried based on key, key range, and composite key queries.

So I think we should be good.

AurelienGasser:
@thibaultrobert @inalgnu I made some changes, please let me know what you think!

1. Gzipping can be turned on or off + add tests
2. Introduce `InitSettings` to configure the chaincode [at instantiation time](https://github.com/SubstraFoundation/hlf-k8s/blob/master/images/hlf-k8s/bin/chaincode-instantiate.sh#L55) + add default values for settings + add tests

So we should be able to enable/disable in hlf-k8s, e.g.

```diff
- peer chaincode instantiate -c '{"Args":["init"]}'
+ peer chaincode instantiate -c '{"Args":["init", "{\"ledger\":{\"enableGzip\":true|false}}"]}'
```
(I haven't tested yet)

The init settings / ledger settings can be expanded in the future to include more stuff.

AurelienGasser:
There's a bug, the substra-tests don't pass.

Error with registernode:

```
400 Client Error: Bad Request for url: http://substra-backend.node-1.com/data_manager/: error while ungzipping data for key MyOrg2MSP gzip: invalid header
```

AurelienGasser:
@inalgnu @thibaultrobert  This is ready for another review. 

I tested with `substra-tests`: it's all :heavy_check_mark:

---

Note: I asked a question [in the HLF chat](https://chat.hyperledger.org/channel/fabric-chaincode-dev?msg=rLKznkpm7D6pgtGpB): 

> In our chaincode, we implemented two different ways to serialize data to be added to the ledger: json and gzipped json. We can control whether or not to gzip assets using a global setting, which is essentially a constant; it doesn't change during the lifecycle of the chaincode. Is there a way to pass arbitrary settings to the chaincode - at install-time for instance? The Init method isn't a good fit because it is not run on every peer. The solution we have now is to hardcode the setting, but that isn't very elegant because the only way to modify the setting is to modify the chaincode source code. 

The only response (so far):

> this is just a guess, but have you tried loading a file from the peer's filesystem? if that works you could just put it in a docker volume and read the file at runtime 

**Edit**: We could inject a config file in `chaincode-install.sh` (in `hlf-k8s`) and read the config file from the `main()` function in the chaincode. I will leave that for another PR!

thibaultrobert (dismissed):
LGTM

AurelienGasser:
@inalgnu Would you min giving this another review?

inelgnu:
Hello @AurelienGasser,
I'm wondering if we are not missing something here, with HLF we have the ability to instantiate multiple chaincodes accessing the same state DB, and I see a limitation for that with this feature. I think the way we instantiate the chaincode should not impact the way the assets are stored in the state DB. 
Shouldn't we use the state DB to store weather the asset is gzipped or not? this should be the case for the entire state DB lifetime. Also, maybe we could add a correspondence table (asset_type | gzipped (true/false)) and rely on this for any chaincode, or an asset struct wrapper {asset asset, gzipped bool}.
Maybe we could take some time to discuss that this afternoon.

Otherwise, If we need this to have a successful load test, I'm fine with merging and take the time to rethink the way we design asset compression afterward.

AurelienGasser:
@inalgnu we could do the following:

1. at chaincode instantiation time, pass a parameter to `Init` describing whether or not to gzip assets. In the peer executing Init, check the state DB for an existing setting already saved:
  - if there's already a setting in the state DB, use that setting and ignore the setting passed to `Init`
  - if there's no setting in the state DB, insert the setting passed to the `Init` function into the state DB
2. In all the peers, on each query/invoke, if the settings haven't been loaded from the state DB, load them.

But I'm not sure having the ability to choose the gzip behavior is desirable: why would we want to turn it on in some ledgers and off in some others? Or on for some assets and off for some others?

AurelienGasser:
Closing until we validate we have a need for this feature

PR #93, opened 2020-02-13
Fix count tuples at creation

thibaultrobert:
At the compute plan creation, let the update compute plan fails if there is only test tuples.


samlesu (approved):
LGTM

PR #94, opened 2020-02-13
Separate ComputePlan state

thibaultrobert:


chaincode/ledger.go
```diff
@@ -191,13 +191,21 @@ type ComputePlan struct {
 	AggregatetupleKeys      []string             `json:"aggregatetupleKeys"`
 	AssetType               AssetType            `json:"assetType"`
 	CompositeTraintupleKeys []string             `json:"compositeTraintupleKeys"`
-	DoneCount               int                  `json:"DoneCount"`
 	IDToTrainTask           map[string]TrainTask `json:"IDToTrainTask"`
-	Status                  string               `json:"status"`
+	State                   ComputePlanState     `json:"-"`
```
AurelienGasser: This might save another developer from doing a google search like I did :p

```suggestion
	State                   ComputePlanState     `json:"-"` // "-" means this field is excluded from JSON (de)serialization
```

chaincode/ledger.go
```diff
@@ -191,13 +191,21 @@ type ComputePlan struct {
...
+	StateKey                string               `json:"stateKey"`
 	Tag                     string               `json:"tag"`
 	TesttupleKeys           []string             `json:"testtupleKeys"`
 	TraintupleKeys          []string             `json:"traintupleKeys"`
-	TupleCount              int                  `json:"tupleCount"`
+}
+
+// ComputePlanState is the ledger's representation of the compute plan state.
+// To minimize the size of every compute plan update it's record under another
```
AurelienGasser: nit

```suggestion
// To minimize the size of every compute plan, update its state record under another
```

chaincode/ledger_db.go
```diff
@@ -93,17 +94,38 @@ func (db *LedgerDB) KeyExists(key string) (bool, error) {
...
 func (db *LedgerDB) Put(key string, object interface{}) error {
 	buff, _ := json.Marshal(object)
 
-	if err := db.cc.PutState(key, buff); err != nil {
+	return db.put(key, buff)
+}
+
+// put stores a value in the chaincode db whether it already exists or not
+func (db *LedgerDB) put(key string, value []byte) error {
+	if err := db.cc.PutState(key, value); err != nil {
```
AurelienGasser: The `put` and `LazyPut` methods add complexity that I think could be avoided by splitting `ComputePlan.Save` into two methods:

```go
// Save the whole compute plan
func (cp *ComputePlan) Save(db *LedgerDB, ID string) error {
	err := db.Put(ID, cp)
	if err != nil {
		return err
	}
	return cp.SaveState(db)
}

// Only save the state
func (cp *ComputePlan) SaveState(db *LedgerDB) error {
	return db.Put(cp.StateKey, cp.State)
}
```

Each caller can then decide if they want to save the whole compute plan - or only the state. This also makes the saving behavior more explicit.

AurelienGasser (changes requested):
Nice work! 

I think there is some extra complexity that can be removed (unless I missed something!) Please let me know what you think of the suggestions below.

chaincode/ledger.go
thibaultrobert: 👍  Indeed that's a good idea.

chaincode/ledger_db.go
thibaultrobert: Well there is different topics here.
First the separation of the compute plan and its state:
The fact that both of them are save distinctively is, from my point of view, a storage optimization and should not be explicit outside this context. Therefore I was not comfortable having a specific method to save the state.
Secondly the `LazyPut`:
I agree it add complexity but I think in the case of big asset like a compute plan it' s a interesting optimization even if we save the state separately. But we can also integrate this mechanism in the ` Put` method and use it for every asset.

AurelienGasser: Re: make `LazyPut` the default. Are there any other cases where we:

- get an asset from the ledger
- then try to save it without modifications ?

I can't think of any but I'm not familiar with all cases. If `LazyPut` would only be beneficial in the case of compute plans, then I'm not sure it's worth adding it since there's a simpler solution available. Moreover, `LazyPut` depends on the state of the cache, so future modifications of the cache system put it at risk of breakage (or worse, data corruption :scream:)

Regarding the two `Save` methods, we could also have a single `Save` method with an argument `stateOnly bool = false`, then we'd still only have a central place for saving that, by default, saves everything, but can be explicitly asked to only save the state. Would that work?

thibaultrobert: You're right stating that it's not a common situation since update are really not the norm. In case of the computeplan it can happen easily and that's the whole reason behind this PR (that and the fact that a single PR should theoretically reach up to 25000 tuples and maybe more in the future) but it's an exception among assets.

So let's drop my previous suggestion. (However I'm not overly worried about the cache since it's re-created at each transaction. The risk of data corruption is really not that high.)

My issue with your proposal `Save` and `SaveState` is not so much the two methods, it's having this separation of the state logic spreading to the tuples' code.  But I'll implement it and we'll see. Let me prove myself wrong.

thibaultrobert:
Ok, it's not much of a change finally. What do you think @AurelienGasser ?

chaincode/compute_plan.go
```diff
@@ -287,8 +287,8 @@ func cancelComputePlan(db *LedgerDB, args []string) (resp outputComputePlan, err
 		return outputComputePlan{}, err
 	}
 
-	computeplan.Status = StatusCanceled
-	err = db.Put(inp.Key, computeplan)
+	computeplan.State.Status = StatusCanceled
+	err = computeplan.Save(db, inp.Key)
```
AurelienGasser:
```suggestion
	err = computeplan.SaveState(db, inp.Key)
```

chaincode/compute_plan.go
```diff
@@ -318,45 +323,50 @@ func (cp *ComputePlan) Save(db *LedgerDB, ID string) error {
 	if err != nil {
 		return err
 	}
-	return nil
+	return db.Put(cp.StateKey, cp.State)
```
AurelienGasser:
```suggestion
	return cp.SaveState(db)
```

chaincode/compute_plan.go
```diff
@@ -318,45 +323,50 @@ func (cp *ComputePlan) Save(db *LedgerDB, ID string) error {
...
+}
+
+// SaveState add or update the compute plan in the ledger
+func (cp *ComputePlan) SaveState(db *LedgerDB, ID string) error {
```
AurelienGasser:
```suggestion
func (cp *ComputePlan) SaveState(db *LedgerDB) error {
```

chaincode/ledger_db.go
```diff
@@ -97,13 +97,18 @@ func (db *LedgerDB) KeyExists(key string) (bool, error) {
 func (db *LedgerDB) Put(key string, object interface{}) error {
 	buff, _ := json.Marshal(object)
 
-	if err := db.cc.PutState(key, buff); err != nil {
+	return db.put(key, buff)
+}
+
+// put stores a value in the chaincode db whether it already exists or not
+func (db *LedgerDB) put(key string, value []byte) error {
```
AurelienGasser: I'm not sure it's necessary to introduce this function - but if you prefer to keep to breakdown, what about renaming `put` to `_PutState` to make the intent more clear?

AurelienGasser (commented):
It looks much simpler now!

AurelienGasser:
I realize you forgot to call `SaveState` from `Save`, unless that's intentional?

AurelienGasser:
Also realized that `ComputePlan.Create` doesn't create the state in the db. Maybe it could call `cp.Save()` (which itself would call `cp.SaveState()`)?

chaincode/ledger_db.go
AurelienGasser: To be clear - what I'm proposing is not to completely separate `Save` and `SaveSate`, it's to have two distinct methods but have `Save` call `SaveState`. This way `SaveState` can be called directly, but `Save` always saves the state too.

AurelienGasser:
`substra-tests` all pass

AurelienGasser (approved):
:+1: :tada:

PR #102, opened 2020-02-26
Return only new ID to Key

thibaultrobert:
We reached a new limit while testing and decided to limit even more the potential size of the compute plan's output. So for now we only return the newly added IDs in the `IDtoKey` field.


AurelienGasser (approved):
LGTM

PR #106, opened 2020-02-28
Intermediate model removal

thibaultrobert:
Companion PR : https://github.com/SubstraFoundation/substra-tests/pull/86 & https://github.com/SubstraFoundation/substra-backend/pull/172 & https://github.com/SubstraFoundation/substra/pull/135


chaincode/ledger_db.go
```diff
@@ -529,6 +529,13 @@ func (db *LedgerDB) AddComputePlanEvent(ComputePlanID, status string) error {
 		return err
 	}
 	cp.AlgoKeys = algokeys
+
+	modelHashes, err := db.GetIndexKeys("model~computePlanID~hash", []string{"model", ComputePlanID})
+	if err != nil {
+		return err
+	}
+	cp.ModelHashes = modelHashes
```
inelgnu: for the largest compute plans I think this field will grow constantly. Maybe we should consider sending only the ones that still to be removed or use another limitation.

thibaultrobert: You right here. It will grow and reach a critical size.
We can go back to the first design suggested for intermediary models: removing at each step the ones that were already used. It will require a bit more logic when handling the compute plan event (we don't want to "clean" an ongoing compute plan).
Or we can only send N models in the event and make the cleaning task call a specific end point to retrieve the next M models to delete. What do you think @inalgnu, @Kelvin-M @AurelienGasser ?

thibaultrobert:
So as decided before we're sending an compute plan event every time an intermediary model can be deleted (when every tuples using this model is done).
There is a potential weakness in the current implementation: the list of models in use in a compute plan is updated on the fly and therefor it's impossible to reconstruct previous event.

chaincode/compute_plan.go
```diff
@@ -404,9 +408,88 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
...
+		// If a tuple has no children it's concidered final and should not be
+		// listed in the index
+		return nil
+	}
+	cp.State.IntermediaryModel = append(cp.State.IntermediaryModel, modelHash)
+
+	return cp.SaveState(db)
+}
+
+// HandleIntermediaryModel is a function
```
AurelienGasser: What would you think of the simpler

```suggestion
// HandleIntermediaryModel ...
```

chaincode/compute_plan.go
```diff
@@ -404,9 +408,88 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
...
+	}
+	if !cp.CleanModels {
+		return nil
+	}
+	allChildKeys, err := db.GetIndexKeys("tuple~inModel~key", []string{"tuple", tupleKey})
+	if err != nil {
+		return err
+	}
+	if len(allChildKeys) == 0 {
+		// If a tuple has no children it's concidered final and should not be
```
AurelienGasser:
```suggestion
		// If a tuple has no children it's considered final and should not be
```

chaincode/compute_plan.go
```diff
@@ -404,9 +408,88 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
...
+	if statusUpdated || len(doneModels) != 0 {
+		db.AddComputePlanEvent(ComputePlanID, cp.State.Status, doneModels)
 		return cp.SaveState(db)
 	}
 	return nil
 }
+
+// ListModelIfIntermediary will reference the hash model if the compute plan ID
+// is not empty and if it's an intermediary model meaning without any children
+func ListModelIfIntermediary(db *LedgerDB, ComputePlanID, tupleKey, modelHash string) error {
```
AurelienGasser: The "list" part of the name makes it look like this function performs a read operation, when it's really a _write_. 

Consider renaming to `TryAddIntermediaryModel`?

chaincode/ledger.go
```diff
@@ -203,9 +204,10 @@ type ComputePlan struct {
 // To minimize the size of every compute plan, update its state record under another
 // key in the ledger. It will reduce the growing rate of the blockchain size.
 type ComputePlanState struct {
-	DoneCount  int    `json:"doneCount"`
-	Status     string `json:"status"`
-	TupleCount int    `json:"tupleCount"`
+	DoneCount         int      `json:"doneCount"`
+	IntermediaryModel []string `json:"intermediaryModel"`
```
AurelienGasser: consider renaming to `IntermediaryModelsInUse` or `ActiveIntermediaryModels` or `PendingIntermediaryModels`?

chaincode/ledger.go
```diff
@@ -190,6 +190,7 @@ type Testtuple struct {
 type ComputePlan struct {
 	AggregatetupleKeys      []string             `json:"aggregatetupleKeys"`
 	AssetType               AssetType            `json:"assetType"`
+	CleanModels             bool                 `json:"cleanModels"`
```
AurelienGasser:
```suggestion
	CleanModels             bool                 `json:"cleanModels"` // whether or not to delete intermediary models
```

chaincode/compute_plan.go
```diff
@@ -404,9 +408,88 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
...
+		// listed in the index
+		return nil
+	}
+	cp.State.IntermediaryModel = append(cp.State.IntermediaryModel, modelHash)
+
+	return cp.SaveState(db)
+}
+
+// HandleIntermediaryModel is a function
+func (cp *ComputePlan) HandleIntermediaryModel(db *LedgerDB) ([]string, error) {
```
AurelienGasser: rename to `UpdateActiveIntermediaryModels`? It would make it clear this updates the list.

chaincode/output.go
```diff
@@ -292,9 +292,10 @@ type Event struct {
 }
 
 type eventComputePlan struct {
+	AlgoKeys      []string `json:"algoKeys"`
 	ComputePlanID string   `json:"computePlanID"`
+	ModelHashes   []string `json:"modelHashes"`
```
AurelienGasser: Consider renaming to `ModelsToDelete`? Currently it's not clear what these model hashes correspond to.

chaincode/compute_plan.go
```diff
@@ -404,9 +408,88 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
...
+		return nil
+	}
+	cp, err := db.GetComputePlan(ComputePlanID)
+	if err != nil {
+		return err
+	}
+	if !cp.CleanModels {
+		return nil
+	}
+	allChildKeys, err := db.GetIndexKeys("tuple~inModel~key", []string{"tuple", tupleKey})
```
AurelienGasser, line 436: Child tuples are considered but not testtuples: is that intentional?

chaincode/input.go
```diff
@@ -157,7 +157,8 @@ type inputComputePlan struct {
 // inputNewComputePlan represent the set of tuples to be added to the compute
 // plan matching the ID
 type inputNewComputePlan struct {
-	Tag string `validate:"omitempty,lte=64" json:"tag"`
+	CleanModels bool   `json:"cleanModels"`
```
AurelienGasser:
```suggestion
	CleanModels bool   `json:"cleanModels"`  // whether or not to delete intermediary models
```

chaincode/compute_plan_test.go
```diff
@@ -162,7 +162,7 @@ func TestModelCompositionComputePlanWorkflow(t *testing.T) {
 	mockStub.MockTransactionStart("42")
 	db := NewLedgerDB(mockStub)
 
-	out, err := createComputePlanInternal(db, modelCompositionComputePlan, tag)
+	out, err := createComputePlanInternal(db, modelCompositionComputePlan, tag, false)
```
AurelienGasser, line 165: Consider adding unit tests for the `cleanModels=true` case?

AurelienGasser (changes requested):
Nice work!!

Testtuples aren't being considered when adding a model to the list: this could be a small bug, maybe?

Regarding naming, `cleanModels` isn't very explicit IMO. TBH I think a better name would be `deleteIntermediaryModels`, but that would require changes in substa/substra-tests/backend :scream: , so maybe not worth the effort?

chaincode/compute_plan.go
thibaultrobert: No that a "I will do it later" and, later on, an overview. Will be fixed.

chaincode/compute_plan.go
thibaultrobert: Yep thanks

chaincode/ledger.go
thibaultrobert: Good idea

thibaultrobert:
> nice work! What about renaming this function to `clearEvents`? "flush" conveys that we send the existing events somewhere which might be misleading.

Done. Well almost: I renamed it `clearEvent` because there is only one `Event` struct.

inelgnu (dismissed):

chaincode/input.go
```diff
@@ -157,7 +157,8 @@ type inputComputePlan struct {
 // inputNewComputePlan represent the set of tuples to be added to the compute
 // plan matching the ID
 type inputNewComputePlan struct {
-	Tag string `validate:"omitempty,lte=64" json:"tag"`
+	CleanModels bool   `json:"cleanModels"` // whether or not to delete intermediary model
```
AurelienGasser:
```suggestion
	CleanModels bool   `json:"cleanModels"` // whether or not to delete intermediary models
```

AurelienGasser (dismissed):
LGTM :+1:  Nice work

inelgnu (approved):
LGTM

PR #108, opened 2020-03-02
Do not update status if testtuple status is the same

inelgnu:


thibaultrobert:
I might missing something but I don't understand in which scenario this can be happening without raising an error ? In a more global rework of our workflow, it will be interesting to challenge the way we log state and the transition rule between those states (do I dare to talk about state machine?). But, IMHO, in the current design we expect an error if we try to pass a tuple to the same state it already is.

thibaultrobert:
Mea culpa,  I've just read the chat conversation about this topic. As I said this errors were by design but I forgot that they were problematic with the first design for cancelling compute plan.

Since the test tuples are now the only one left with this mechanism, let's removed it. But I think that it should be shown somewhere in documentation that such a transition is possible.

thibaultrobert (approved):

PR #116, opened 2020-05-15
Add CP ID to tuples hashes

thibaultrobert:
When creating a big compute plan, if a `update_compute_plan` step failed the user can not try to recreate the compute plan from scratch because some of its training task have already been created.
To mitigate this issue, a quick fix is implemented here: we include the compute plan ID of a tuple as in his hash therefore two identical tuples can be created in two different compute plan.


AurelienGasser (approved):
LGTM !

(assuming e2e tests pass!)

thibaultrobert:
End to end tests are ok.

PR #118, opened 2020-05-22
Make sure to set CP ID before hashing

### LINKED ISSUE #117: [BUG] Cannot create duplicate composite traintuples in compute plan
It's not possible to use the same composite traintuple in multiple compute plans using `updateComputePlan`. 

I suspect the issue also applies to other types of training tasks (traintuples, etc)

**To Reproduce**

- Create a compute plan
- Add a composite traintuple using `updateComputePlan`
- Create another compute plan
- Add the same composite traintuple  using `updateComputePlan`

**Expected behavior**

No error, a new composite traintuple is created (duplicate)

**Actual behavior**

```
Requests error status 400: {"message":"traintuple ID 9cf92be7f911494bb38ca8499dfeb033: composite traintuple already exists","computePlanID":"f9a3561f3139b237cc5675edeb7172d88a1a17204b56324588ebbed36ae09f0e"}
```

**Environment:**

master (https://github.com/SubstraFoundation/substra-chaincode/commit/bd634f31459fd2b456d6a93ac349093759dd9ba4)

**Additional context**

I suspect the issue comes from the fact that `compositeTraintuple.ComputePlanId` is set inside `AddToComputePlan`, which is called _after_ `getKey`. I gave this very little research so I could be completely off mark here ;)

Considering that adding duplicate training/testing tasks is an important use-case for melloddy, I would argue it would be worth adding unit tests for traintuples, compositetraintuples, aggregatetuples and testtuples.

thibaultrobert:
Should fix #117

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -1432,8 +1432,8 @@ peer chaincode invoke -n mycc -c '{"Args":["createComputePlan","{\"cleanModels\"
 ```json
 {
  "IDToKey": {
-  "firstTraintupleID": "a3cbf90bf2945cc7728f27894acf8e609fd65ef93340157f90706bd441d5f1b8",
-  "secondTraintupleID": "61f210813e18f3a7af124ff32d6f5a46a276d5ae9fdd5d5c6f49a8e49c88366d"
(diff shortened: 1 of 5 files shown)


AurelienGasser (approved):
LGTM

thibaultrobert:
E2e ok by the way

inelgnu (approved):

PR #123, opened 2020-07-28
Check done intermediary models only when tuple is Done

inelgnu:
This will perform a check of done intermediary models only when a tuple is Done, to avoid MVCC.

This PR is targeting master, the commit is already present is melloddy branch and the tag 0.0.9-melloddy.1

### DIFF
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -414,7 +414,11 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
 		return err
 	}
 	statusUpdated := cp.UpdateStatus(tupleStatus)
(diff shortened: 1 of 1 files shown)


thibaultrobert (dismissed):
Becarefull the size of the event that this update generate will grow with the size of the compute plan and may exceed GRPC size limit.

inelgnu:
@thibaultrobert I don't understand your point here, why would this impact the new behavior and not the previous one?

thibaultrobert:
https://github.com/SubstraFoundation/substra-chaincode/pull/123#issuecomment-665272500
It won't you're right. I've completely misunderstood your commit here. Sorry about that and once again the noise.

AurelienGasser:
End to end tests: :heavy_check_mark:

AurelienGasser (approved):

PR #124, opened 2020-07-28
Intermediate models: Refactor / renaming

AurelienGasser:
Rename some functions/variables + modify the flow slightly, for extra clarity.

### DIFF
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -346,7 +346,7 @@ func (cp *ComputePlan) SaveState(db *LedgerDB) error {
 	return db.Put(cp.StateKey, cp.State)
 }
 
-// UpdateStatus check the tuple status (from an updated tuple or a new one)
+// UpdateState check the tuple status (from an updated tuple or a new one)
 // and, if required, it updates the compute plan' status and/or its doneCount.
 // It returns true if there is any change to the compute plan, false otherwise.
 func (cp *ComputePlan) UpdateState(tupleStatus string) bool {
--- a/chaincode/compute_plan.go
+++ b/chaincode/compute_plan.go
@@ -417,7 +417,6 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
 	doneModels := []string{}
 	if tupleStatus == StatusDone && cp.CleanModels {
 		doneModels, err = cp.UpdateIntermediaryModelsInuse(db)
-		stateUpdated = true
 	}
 
 	if err != nil {
@@ -428,8 +427,6 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
 		if err != nil {
 			return err
 		}
-	}
-	if stateUpdated {
 		return cp.SaveState(db)
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


inelgnu (approved):

thibaultrobert (dismissed):
Beware the event compute plan are not only use at the end of one compute plan but also to delete intermediary models. That the reason events are send for each status update

thibaultrobert (commented):
> In this case, `UpdateStatus` should return false, and consequently no CP event will be generated.﻿


Reading this I though you only send event at the end of the compute plan, which is not what you're saying nor what this PR does. Sorry about that.
On the other hand this PR may cause a bug because the output of `UpdateStatus` not only decide if there is an event sent but also if the state of the CP should be saved (see [here](https://github.com/SubstraFoundation/substra-chaincode/blob/89b489257e22c79dc1905136295e8cafac45484a/chaincode/compute_plan.go#L430)).

AurelienGasser:
@thibaultrobert Good catch, thanks!! 

I changed the purpose of this PR: instead of changing the behavior, it just renames some stuff and changes the flow slightly to make it easier to understand. 

Would you mind giving it a review?

chaincode/compute_plan.go
```diff
@@ -413,20 +413,23 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
 	if err != nil {
 		return err
 	}
-	statusUpdated := cp.UpdateStatus(tupleStatus)
+	stateUpdated := cp.UpdateState(tupleStatus)
 	doneModels := []string{}
-	if tupleStatus == StatusDone {
-		doneModels, err = cp.CheckDoneIntermediaryModel(db)
+	if tupleStatus == StatusDone && cp.CleanModels {
```
thibaultrobert, line 418: 👍

thibaultrobert: It's clearer but it's also checked at the fist line of `CheckDoneIntermediaryModel`/`UpdateIntermediaryModelsInuse`

chaincode/compute_plan.go
```diff
@@ -413,20 +413,23 @@ func UpdateComputePlanState(db *LedgerDB, ComputePlanID, tupleStatus, tupleKey s
...
 		return err
 	}
-	if statusUpdated || len(doneModels) != 0 {
+	if stateUpdated || len(doneModels) != 0 {
 		err = db.AddComputePlanEvent(ComputePlanID, cp.State.Status, doneModels)
 		if err != nil {
 			return err
 		}
+	}
+	if stateUpdated {
```
thibaultrobert: Why not saving the state in the case the intemediary model list has been updated? Because if `doneModels` is not empty, then the list of _models in use_ present in the compute plan state has also been updated (cf [here](https://github.com/SubstraFoundation/substra-chaincode/blob/89b489257e22c79dc1905136295e8cafac45484a/chaincode/compute_plan.go#L508)) and therefore this compute plan state need to be saved.

chaincode/compute_plan.go
AurelienGasser: This extra check is not all that useful, I could remove it

chaincode/compute_plan.go
AurelienGasser: I agree, let me push an update.

chaincode/compute_plan.go
thibaultrobert: Yes but at first read I thought it was useful even though I've wrote the other check myself... So as I said it's not useful but probably clearer. Therefore I've no strong opinion either way.

thibaultrobert (approved):

AurelienGasser:
@thibaultrobert thanks for the review!

thibaultrobert:
Kein Problem, gute Nacht.

PR #131, opened 2020-11-26
Chaincode Pagination with query[Asset]s.

Kelvin-M:
## Description
The aim of this PR is to add chaincode pagination for query[Asset]s.
It's based on :
- https://hyperledger-fabric.readthedocs.io/en/release-2.2/couchdb_tutorial.html#cdb-pagination
- https://godoc.org/github.com/hyperledger/fabric-chaincode-go/shim#ChaincodeStub.GetStateByPartialCompositeKey
- https://godoc.org/github.com/hyperledger/fabric-chaincode-go/shim#ChaincodeStub.GetStateByPartialCompositeKeyWithPagination

## Closes issue(s)
None

## Companion PRs
https://github.com/SubstraFoundation/substra-backend/pull/352

## How to test / repro
Launch e2e tests, no regression should appear.

## Screenshots / Trace
None

## Changes include
- [x] Use `GetStateByPartialCompositeKeyWithPagination` instead of `GetStateByPartialCompositeKey`
- [x] Return bookmark to the user and and bookmark args
- [x] Update tests

## Checklist
- [x] I have tested this code
- [x] I have updated examples
- [x] Restore old limit value (5 -> 500)

### DIFF
--- a/chaincode/algo.go
+++ b/chaincode/algo.go
@@ -93,10 +93,9 @@ func queryAlgo(db *LedgerDB, args []string) (out outputAlgo, err error) {
 }
 
 // queryAlgos returns all algos of the ledger
-func queryAlgos(db *LedgerDB, args []string) (outAlgos []outputAlgo, bookmarks map[string]string, err error) {
+func queryAlgos(db *LedgerDB, args []string) (outAlgos []outputAlgo, bookmark string, err error) {
+	inp := inputBookmark{}
 	outAlgos = []outputAlgo{}
-	index := "algo~owner~key"
-	bookmarks = map[string]string{index: ""}
 
 	if len(args) > 1 {
 		err = errors.BadRequest("incorrect number of arguments, expecting at most one argument")
@@ -104,16 +103,13 @@ func queryAlgos(db *LedgerDB, args []string) (outAlgos []outputAlgo, bookmarks m
 	}
 
 	if len(args) == 1 && args[0] != "" {
-		inp := inputBookmarks{}
-		err := AssetFromJSON(args, &inp)
+		err = AssetFromJSON(args, &inp)
 		if err != nil {
-			return nil, bookmarks, err
+			return
 		}
-		bookmarks = inp.Bookmarks
 	}
 
-	elementsKeys, bookmark, err := db.GetIndexKeysWithPagination(index, []string{"algo"}, OutputAssetPaginationHardLimit, bookmarks[index])
-	bookmarks[index] = bookmark
+	elementsKeys, bookmark, err := db.GetIndexKeysWithPagination("algo~owner~key", []string{"algo"}, OutputPageSize, inp.Bookmark)
 
 	if err != nil {
 		return
@@ -122,7 +118,7 @@ func queryAlgos(db *LedgerDB, args []string) (outAlgos []outputAlgo, bookmarks m
 	for _, key := range elementsKeys {
 		algo, err := db.GetAlgo(key)
 		if err != nil {
-			return outAlgos, bookmarks, err
+			return outAlgos, bookmark, err
 		}
 		var out outputAlgo
 		out.Fill(algo)
--- a/chaincode/algo.go
+++ b/chaincode/algo.go
@@ -93,21 +93,28 @@ func queryAlgo(db *LedgerDB, args []string) (out outputAlgo, err error) {
 }
 
 // queryAlgos returns all algos of the ledger
-func queryAlgos(db *LedgerDB, args []string) (outAlgos []outputAlgo, err error) {
+func queryAlgos(db *LedgerDB, args []string) (outAlgos []outputAlgo, bookmark string, err error) {
 	outAlgos = []outputAlgo{}
-	if len(args) != 0 {
-		err = errors.BadRequest("incorrect number of arguments, expecting nothing")
+
+	if len(args) > 1 {
+		err = errors.BadRequest("incorrect number of arguments, expecting at most one argument")
 		return
 	}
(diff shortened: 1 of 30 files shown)


AurelienGasser:
@Kelvin-M I moved to a simpler `string` bookmark structure. 

For the `getModels` special case, I used a JSON-ified structure.

I also made sure smart contracts that support bookmarks _always_ return a bookmark in the chaincode response:

```go
{
result interface{}
bookmark string
}
```

even when the bookmark is `""`. This way the return type is consistent accross calls.

Lastly I added 2 unit tests:

- one for traintuples (basic case)
- one for getModels (specia case)

I had to make changes to the mock stub to support pagination.

Please let me know if all that looks good to you.

AurelienGasser (approved):

chaincode/algo.go
```diff
@@ -93,21 +93,32 @@ func queryAlgo(db *LedgerDB, args []string) (out outputAlgo, err error) {
...
-	if len(args) != 0 {
-		err = errors.BadRequest("incorrect number of arguments, expecting nothing")
+
+	if len(args) > 1 {
+		err = errors.BadRequest("incorrect number of arguments, expecting at most one argument")
 		return
 	}
-	elementsKeys, err := db.GetIndexKeys("algo~owner~key", []string{"algo"})
+
+	if len(args) == 1 && args[0] != "" {
```
thibaultrobert, line 105: Ok correct me then ignore the rest if I'm mistaken but you want the following behaviors: 
-  a call with no arg should work as before (so with `inp.Bookmark = ""`)
-  a call with arg should only have one json as first argument and it should be compatible with the `inputBookmark` struct

If so I would only test if there is args (and eventually if it's not an empty string but I'm not sure why) because `AssetFromJSON` already raise an error if there is more arg. Something like that (with the empty string check):
```go
if len(args) > 1 && arg[0] != "" {
```

Of course the drawback here will be the error requesting an exact number of argument rather than a maximum. But as this pagination is only use automatically by the back-end, I don't think a clearer error here is required.

chaincode/algo_aggregate.go
```diff
@@ -93,21 +93,32 @@ func queryAggregateAlgo(db *LedgerDB, args []string) (out outputAggregateAlgo, e
...
-	if len(args) != 0 {
-		err = errors.BadRequest("incorrect number of arguments, expecting nothing")
+
+	if len(args) > 1 {
+		err = errors.BadRequest("incorrect number of arguments, expecting at most one argument")
 		return
 	}
-	elementsKeys, err := db.GetIndexKeys("aggregateAlgo~owner~key", []string{"aggregateAlgo"})
+
+	if len(args) == 1 && args[0] != "" {
```
thibaultrobert, line 105: Same here and others following that I will not comment on.

chaincode/main.go
```diff
@@ -189,17 +204,25 @@ func (t *SubstraChaincode) Invoke(stub shim.ChaincodeStubInterface) peer.Respons
...
 	}
 
+	// Add bookmark (if any) to response
+	if hasBookmark {
+		result = map[string]interface{}{
+			"results":  result,
+			"bookmark": bookmark,
+		}
+	}
+
```
thibaultrobert, line 220: I'm not a fan of this format but I understand the need. 
And I feel that this mechanic should be handle through interface for lists type output. But one again this is more complex and can come later.

chaincode/ledger_db.go
```diff
@@ -176,6 +177,44 @@ func (db *LedgerDB) GetIndexKeys(index string, attributes []string) ([]string, e
 	return keys, nil
 }
 
+// GetIndexKeysWithPagination returns keys matching composite key values from the chaincode db
+func (db *LedgerDB) GetIndexKeysWithPagination(index string, attributes []string, pageSize int32, bookmark string) ([]string, string, error) {
```
thibaultrobert, line 181: May be minor but why passing pageSize as a parameter if you always use the same global.

I would propose to pass it during the initialization of the `LedgerDB` to make it explicit enough but still limit the number of parameters to a minimum.

thibaultrobert (commented):
A few feedback but feel free to ignore them.

Nice to see the start of pagination.

chaincode/ledger_db.go
Kelvin-M: Not exactly.
If you look [here](https://github.com/SubstraFoundation/substra-chaincode/pull/131/files#diff-95f619a253f6926eaeb3deb37d0536e95f0418a7ecdb444145fb2080bfecfcf5R133) you can see that we not always use the same global everywhere. For this case, as we do multiple calls, we need to divide it by 3.

chaincode/main.go
Kelvin-M: As it's only the first step of full pagination, we will see late if we can have a better format at the end :)

chaincode/algo.go
Kelvin-M: As it's the first step of full pagination in get/list asset(s) chaincode response, we don't already know if we will send the response to the user or not. Better to make it explicit now I think !

chaincode/ledger_db.go
thibaultrobert: Ok.

chaincode/algo.go
thibaultrobert: I'm not sure I agree on this because for an end user pagination I think you'll change other thing as well and it will not be a wrapped response anymore. But it's not a big issue so you can keep it that way. I just find it less explicit when reading the code.

thibaultrobert (approved):

PR #132, opened 2020-12-03
Remove vendor folder as it is not used

### LINKED ISSUE #25: Use vendoring?
Currently, we have a `vendor/` directory, which indicates that we want to use vendored dependencies. However, **the project doesn't currently use vendored dependencies consistently**. 

In particular:

1. Some of the dependencies are missing from the `vendor/` directory
    - You can see it by typing `go build -mod=vendor`: this command is going to complain about missing dependencies
    - The fix would be to run `go mod vendor` and commit the updated vendored dependencies [1]
2. The README doesn't instruct users to use the `-mod=vendor` option. Without this option, the vendored dependencies are ignored [2]
3. The CI build doesn't use vendored dependencies (it doesn't use `-mod=vendor`)
4. The CI build doesn't check the integrity of dependencies (it doesn't call `go mod verify`) [3]
5. I'm not sure whether `hlf-k8s` uses vendored dependencies when instanciating the chaincode?

Hence my question:  **Do we want to use vendored dependencies?** 

If so, in order to be consistent, we should address the points above :)

[1] https://tip.golang.org/cmd/go/#hdr-Make_vendored_copy_of_dependencies

[2] https://tip.golang.org/cmd/go/#hdr-Modules_and_vendoring

[3] https://tip.golang.org/cmd/go/#hdr-Verify_dependencies_have_expected_content

Kelvin-M:
Fix #25


thibaultrobert (approved):
LGTM (check comment [here](https://github.com/SubstraFoundation/substra-chaincode/issues/25#issuecomment-738075351))

AurelienGasser (approved):

PR #133, opened 2020-12-08
standalone chaincode

AurelienGasser:
Companion PR: 

- https://github.com/SubstraFoundation/substra-backend/pull/356

Missing (non exhaustive list):

- [x] Thread safety (allow multiple requests in parallel)
- [x] backend: Wait for migrations to be finished before starting to listen to events
- [x] Multiple MSPs (right now everybody is MyOrg1MSP)
- [ ] Caller verification (make sure you can't impersonate another org)
- [ ] CouchDB (right now everything is in memory)
- [ ] option to start in standalone / hlf-mode
- ...


chaincode/main.go
```diff
@@ -223,7 +233,7 @@ func (t *SubstraChaincode) Invoke(stub shim.ChaincodeStubInterface) peer.Respons
 
 	if err != nil {
 		logger.Infof("[%s][%s] Response (%dms): '%#v','%s'", stub.GetChannelID(), stub.GetTxID()[:10], duration, result, bookmark)
-		return formatErrorResponse(errors.Internal("could not format response: %s", err.Error()))
```
AurelienGasser, line 226: Todo: fix

chaincode/main.go
```diff
@@ -49,7 +49,15 @@ func (t *SubstraChaincode) Init(stub shim.ChaincodeStubInterface) peer.Response
...
 func (t *SubstraChaincode) Invoke(stub shim.ChaincodeStubInterface) peer.Response {
+	resp, _, err := t._Invoke(stub)
 
+	if err != nil {
+		return formatErrorResponse(err)
+	}
+	return shim.Success(resp)
+}
+
+func (t *SubstraChaincode) _Invoke(stub shim.ChaincodeStubInterface) ([]byte, *Event, error) {
```
inalgnu, line 60: [CS] Go private function should be lower case.
- CapitalCase function means package exported function
- lowerCase means package internal function
no need to prefix with _ is not idiomatic go

AurelienGasser: Oh yeah, this is very very WIP, I didn't think too much, I was just trying to make things work.

Speaking of exporting functions, I tried to move the code related to the standalone functionality to a new package named "standalone", but this doesn't work: I can't reference anything from the "main" package inside the "standalone" package. 

I was thinking we could create a "chaincode" package and move everything currently in main (except the actual main function) there. This could possibly be done prior to and independently from this PR. 

Then :

- We should be in a better place to introduce a "standalone" package
- over time (in future PRs) we can slowly extract meaningful packages out of the "chaincode" package

AurelienGasser: Or we can just keep stuffing stuff in main (like I did in this PR)...

Kelvin-M:
It can be closed now no ?

AurelienGasser:
@Kelvin-M yes, but we should keep the branch for now (until an alternate implementation is close to ready).

Closing.

PR #134, opened 2020-12-14
migrate from Travis CI to GitHub Actions

AurelienGasser:
It's all in the title!

Builds should be triggered for

- pushes to master
- pushes to PRs (regardless of target branch)

Build times are comparable as with Travis (~1 min)

image

Once we move to a more recent version of Go, the "Install Go" step should be faster (version 1.12 is not cached)


.github/workflows/go.yml
```diff
@@ -0,0 +1,34 @@
+name: Build and test Go
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: "*"
```
AlexandrePicosson: Here you can target only master, this has the benefit of running the CI only once on the code on a scenario like this:
```
master ---------------
                 /
pr1 ------------/
         /
pr2 ----/
```

With your solution you test on `pr2 -> pr1` and `pr1 -> master` and changing `branches:` to `master` you would build only on `pr1 -> master`.

.github/workflows/go.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          path: ~/go/pkg/mod
+          key: build-${{ hashFiles('**/go.sum') }}
+      - name: Build
+        run: go build -v
+        working-directory: ./chaincode
+      - name: Test
+        run: go test -v ./...
+        working-directory: ./chaincode
+      - name: Lint
```
AlexandrePicosson: You can run the lint first in this testing pipeline, that way we fail early without running the potentially costly operation of building the go package (build takes `13s` vs `2s` for the lint) if the code does not comply with the lint rules.

.github/workflows/go.yml
AurelienGasser: This actually intentional. Unit testing `pr2 -> pr1` is also useful, to catch errors early, no?

AlexandrePicosson: This was not done previously with travis but if this is a feature you would like to have then we can revert the change and put it back like it was before.

AlexandrePicosson (approved):

mblottiere (approved):

.github/workflows/go.yml
```diff
@@ -0,0 +1,37 @@
...
+  build:
+    name: Build and test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.12.x
```
inelgnu, line 22: Maybe not related to this PR, but I think we should consider using a more recent version of Go.

inelgnu (commented):
LGTM, minor comment on version that could be treated seperately

inelgnu (approved):

.github/workflows/go.yml
AurelienGasser: Agreed. I used 1.12 as this is the documented/tested go version for the chaincode (in the README and go.mod), but we should move to 1.15.

PR #135, opened 2021-01-08
[permissions] Do not fetch node on public permissions

mblottiere:
This patch keeps the existing behavior while updating the logic so
that the nodes are only checked when needed.
Basically, instead of iterating over nodeIds and always skipping, it
now only iterate when Process.Public is false.

### DIFF
--- a/chaincode/permissions.go
+++ b/chaincode/permissions.go
@@ -53,25 +53,9 @@ func (perms Permissions) CanProcess(owner, node string) bool {
 
 // NewPermissions create the Permissions according to the arg received
 func NewPermissions(db *LedgerDB, in inputPermissions) (Permissions, error) {
-	nodes, err := queryNodes(db, []string{})
-	if err != nil {
-		return Permissions{}, err
-	}
-
-	nodesIDs := []string{}
-	for _, node := range nodes {
-		nodesIDs = append(nodesIDs, node.ID)
-	}
-
-	// Validate Process inputPermissions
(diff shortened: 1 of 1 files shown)


chaincode/permissions.go
```diff
@@ -53,25 +53,23 @@ func (perms Permissions) CanProcess(owner, node string) bool {
...
-	nodes, err := queryNodes(db, []string{})
-	if err != nil {
-		return Permissions{}, err
-	}
-
-	nodesIDs := []string{}
-	for _, node := range nodes {
-		nodesIDs = append(nodesIDs, node.ID)
-	}
+	if !in.Process.Public {
```
AurelienGasser, line 56: What would you think of the following to reduce indentation

```
if in.Process.Public {
  return
}
```

**Edit**

or to be more explicit `return Permissions{}, nil`

mblottiere: Hmm we would still need to create the Permission object to assign proper owner and process/download perms.
That would mean moving this part before the validation, which I'd rather not do since this is a loaded footgun to bypass the validation.
I will extract this check in another function to cleanly separate the concerns.

AurelienGasser: @mblottiere Oh my bad, I misread the function, your initial implementation is great.

inelgnu (approved):
Good catch !

AurelienGasser (approved):

PR #138, opened 2021-02-01
Reduce MVCC conflicts during compute plans

AurelienGasser:
Changes:

- Break down compute plan state into multiple sub-states (one per organization).
- Check if intermediary models are ready for deletion in a more concurrency-friendly way 
  - Before: when a tuple becomes "done", check if its parent has all its children in the "done" state
  - After: when a new model is created, look at the last models created by the same worker, and check if they're ready for deletion

These changes collectively significantly reduce  the number of MVCC conflicts occurring in compute plan scenarios involving many organizations (90%+ reduction in some cases).

This PR also introduces changes in unit tests which using multiple workers possible. Previously, all the smart contracts were invoked as `SampleOrg`. Now certain compute plan-related unit tests invoke smart contracts using multiple workers, simulating a real federated learning flow.


AurelienGasser:
cc @Kelvin-M @mblottiere @AlexandrePicosson FYI

chaincode/compute_plan_worker_state.go
```diff
@@ -0,0 +1,194 @@
...
+		if inUse {
+			usedModels = append(usedModels, modelKey)
+		} else {
+			unusedModels = append(unusedModels, modelKey)
+		}
+	}
+	return usedModels, unusedModels, nil
+}
+
+// isModelInuse returns true if the model with the supplied key is the
```
mblottiere:
```suggestion
// isModelInUse returns true if the model with the supplied key is the
```

chaincode/compute_plan_worker_state.go
```diff
@@ -0,0 +1,194 @@
...
+		if err != nil {
+			return []string{}, err
+		}
+	}
+	return res, nil
+}
+
+// getModelsInuse takes a list of model keys and checks whether these models are still in use or not.
+// It returns the initial list split into two sublists: the models in use, and the models unused.
+func getModelsInuse(db *LedgerDB, modelKeys []string) (usedModels []string, unusedModels []string, err error) {
```
mblottiere:
```suggestion
func getModelsInUse(db *LedgerDB, modelKeys []string) (usedModels []string, unusedModels []string, err error) {
```

Kelvin-M (approved):
LGTM after some tests ! 
Did you try on a big compute plan to see if intermediary models are deleted ?

chaincode/compute_plan.go
```diff
@@ -367,46 +389,68 @@ func (cp *ComputePlan) SaveState(db *LedgerDB) error {
 // UpdateState check the tuple status (from an updated tuple or a new one)
 // and, if required, it updates the compute plan' status and/or its doneCount.
 // It returns true if there is any change to the compute plan, false otherwise.
```
mblottiere:
```suggestion
// It returns true and the list of models to delete if there is any change to the compute plan, false and empty list otherwise.
```

chaincode/compute_plan_worker_state.go
```diff
@@ -0,0 +1,194 @@
...
+			return false, nil
+		}
+		if tuple.Status != StatusDone {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// getTupleChildren returns the keys of all the tuple which have the supplied tuple as an in-model
```
mblottiere:
```suggestion
// getTupleChildren returns the keys of all the tuples which have the supplied tuple as an in-model
```

chaincode/compute_plan_worker_state.go
```diff
@@ -0,0 +1,194 @@
...
+// If includeTesttuples is True, return the testtuple children, else omit them,
```
mblottiere:
```suggestion
// If includeTesttuples is True, also include the testtuple children, else omit them.
```

chaincode/compute_plan_worker_state.go
```diff
@@ -0,0 +1,194 @@
+package main
+
+import "fmt"
+
+// TryAddIntermediaryModel will reference the model key if the compute plan key
```
mblottiere: What is the meaning of `reference` here?
There is a lot going on: we update the worker state, emit a compute plan event.
This documentation should be more explicit.

mblottiere (approved):
Great work 🚀

AurelienGasser:
> Did you try on a big compute plan to see if intermediary models are deleted ?

Yes

chaincode/compute_plan_worker_state.go
AurelienGasser: Good catch! This comment was copy pasted from the previous implementation. I'll update it.

chaincode/compute_plan_worker_state.go
```diff
@@ -0,0 +1,194 @@
...
+		return false, err
+	}
+	if len(keys) == 0 {
+		// This occurs for the keys added during the same transaction. But
+		// thoses models can just be considered in use
+		return true, nil
+	}
+	children, err := getTupleChildren(db, keys[0], true)
+	if err != nil {
+		return false, nil
```
inelgnu: why you are not handling this error (same on line 168)? is it intentional?

chaincode/ledger.go
```diff
@@ -215,16 +216,22 @@ type ComputePlan struct {
...
 type ComputePlanState struct {
-	DoneCount               int      `json:"done_count"`
+	Status string `json:"status"`
+}
+
+// ComputePlanWorkerState contains state information for a given
+// compute plan and worker
+type ComputePlanWorkerState struct {
 	IntermediaryModelsInUse []string `json:"intermediary_models_in_use"`
-	Status                  string   `json:"status"`
```
inelgnu, line 226: this looks as a breaking change. maybe we could keep it for now (or deprecated first)? is the backend updated also?

chaincode/compute_plan_worker_state.go
AurelienGasser: Not intentional, it's a mistake

AurelienGasser: fixed in https://github.com/SubstraFoundation/substra-chaincode/pull/138/commits/247efff256282ad69bded0a46c7f99ed422b1d0b

chaincode/compute_plan_worker_state.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// getTupleCounts returns the number of tuples in the "done" state and the total number of tuples
+// for a given compute plan
+func (cp *ComputePlan) getTupleCounts(db *LedgerDB) (doneCount int, tupleCount int, err error) {
+	for _, worker := range cp.Workers {
+		wStateKey := cp.getCPWorkerStateKey(worker)
+		wState, err := db.GetCPWorkerState(wStateKey)
+		if err != nil {
+			return doneCount, tupleCount, nil
```
inelgnu: looks like you  are not returning an error if there is one

chaincode/compute_plan_worker_state.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+		doneCount += wState.DoneCount
+		tupleCount += wState.TupleCount
+	}
+	return doneCount, tupleCount, nil
+}
+
+// getCPWorkerStateKey returns the worker state key for a given compute plan and worker
+func (cp *ComputePlan) getCPWorkerStateKey(worker string) string {
+	return fmt.Sprintf("computePlan~%v~doneCountByWorker~%v", cp.Key, worker)
```
inelgnu: the function name is bit confusing (maybe I'm wrong) but look it's based on doneCountByWork only. Does it makes sens to call it getCPWorkerDoneKey?

chaincode/ledger.go
AurelienGasser: There's no update necessary in the backend: to the backend the change is transparent.

Re: breaking. I ran a test. I started two compute plans with the current `master` version of the chaincode. 

image


Once the CPs were done I upgraded the chaincode to this branch. This is what I see in `substra list compute_plan`

image


So we lose the counts, but we can safely "get" compute plans without an error. 

I haven't tested adding a new tuple to the existing CP. In that case we would have undetermined behavior. Handling full backwards compatibility would be much more work, as it would involve writing a full migration.

chaincode/compute_plan_worker_state.go
AurelienGasser: fixed in https://github.com/SubstraFoundation/substra-chaincode/pull/138/commits/e833fe06b2f2002848fb1b8a66b89c59bc18ff3d

chaincode/compute_plan_worker_state.go
AurelienGasser: fixed in https://github.com/SubstraFoundation/substra-chaincode/pull/138/commits/e833fe06b2f2002848fb1b8a66b89c59bc18ff3d

chaincode/ledger.go
inelgnu: ok, let's keep things simple for now and tackle the migration process in another track.

We need at least make sure that it is mentioned in the next release and update Substra to handle the status properly + update compatibility table as usual.

inelgnu (commented):
Nice work @AurelienGasser, kudos for offering this solution that will speed up the runs 👏

inelgnu (approved):

PR #139, opened 2021-03-18
fix: incorrect model permissions

AurelienGasser:
**SECURITY FIX**

Companion PR: https://github.com/SubstraFoundation/substra-tests/pull/179

The return value of `queryModelPermissions` was (incorrectly) "public" in all cases, except for head composite traintuple models. Effectively, this allowed unrestricted node-to-node model transfers, regardless of dataset/algo permissions.

This PR fixes permissions in all other cases:

- traintuple out-models
- aggregatetuple out-models
- composite traintuple trunk models

A non-regression test is also added.

### DIFF
--- a/chaincode/input_test.go
+++ b/chaincode/input_test.go
@@ -21,6 +21,16 @@ var (
 )
 
 func (dataManager *inputDataManager) createDefault() [][]byte {
+	dataManager.fillDefaults()
+	return dataManager.getArgs()
+}
+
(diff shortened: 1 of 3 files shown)


mblottiere (approved):

Kelvin-M:
@camillemarini  @thibaultrobert  @samlesu @inalgnu Can you review as it's a security fix !

samlesu (approved):

PR #140, opened 2021-03-26
fix: fix inplace sort to compare datasample keys

Kelvin-M:
Inplace sort to compare datasample keys break the usage with user datasample order when register testtuple
Companion PR : https://github.com/SubstraFoundation/substra-tests/pull/181

### DIFF
--- a/chaincode/testtuple.go
+++ b/chaincode/testtuple.go
@@ -21,6 +21,23 @@ import (
 	"strconv"
 )
 
+func isEqual(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	aCopy := make([]string, len(a))
+	bCopy := make([]string, len(b))
+
+	copy(aCopy, a)
+	copy(bCopy, b)
+
+	sort.Strings(aCopy)
+	sort.Strings(bCopy)
+
+	return reflect.DeepEqual(aCopy, bCopy)
+}
+
 // -------------------------------------------------------------------------------------------
 // Methods on receivers testtuple
 // -------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Kelvin-M:
E2E tests pass

mblottiere (approved):

chaincode/testtuple.go
```diff
@@ -21,6 +21,23 @@ import (
...
+	bCopy := make([]string, len(b))
+
+	copy(aCopy, a)
+	copy(bCopy, b)
+
+	sort.Strings(aCopy)
+	sort.Strings(bCopy)
+
+	return reflect.DeepEqual(aCopy, bCopy)
+}
```
AurelienGasser, line 39: is there a reason not to use a simple for-loop here?

chaincode/testtuple.go
```diff
@@ -72,8 +86,7 @@ func (testtuple *Testtuple) SetFromInput(db *LedgerDB, inp inputTesttuple) error
 			return err
 		}
 		dataManagerKey = inp.DataManagerKey
-		sort.Strings(dataSampleKeys)
-		testtuple.Certified = objectiveDataManagerKey == dataManagerKey && reflect.DeepEqual(objectiveDataSampleKeys, dataSampleKeys)
+		testtuple.Certified = objectiveDataManagerKey == dataManagerKey && isEqual(objectiveDataSampleKeys, dataSampleKeys)
```
AurelienGasser, line 89: For future readers it would be helpful to explain (in a comment?) that the `isEqual` function ensures the order is the same (and not only that the elements are the same, possibly unsorted)

AurelienGasser (changes requested):

chaincode/testtuple.go
AurelienGasser:
```go
func isEqual(a []string, b []string) bool {
    if len(a) != len(b) {
        return false
    }
    for i, v := range a {
        if v != b[i] {
            return false
        }
    }
    return true
}
```

AurelienGasser: Please ignore this comment, misunderstanding on my part

chaincode/testtuple.go
AurelienGasser: Please ignore this comment, misunderstanding on my part

AurelienGasser (approved):

PR #142, opened 2021-05-01
fix: aggregate permissions

AurelienGasser:
This PR fixes incorrect permissions for aggregate models. 

Scenario:

- N composite traintuples on N different worker nodes. Each of these composite traintuples only grants process permission to itself.
- 1 aggregatetuple, which aggregates the N composite traintuples

```
composite 1: {worker: "node_1", permissions: {public: false, authorized_ids: ["node_1"]}}
...
composite N: {worker: "node_N", permissions: {public: false, authorized_ids: ["node_N"]}}
``` 

### Before this PR:

The aggregate permission is the intersection of parent permissions. In this case, **nobody** has process permissions.

```
aggregate: {permissions: {public: false, authorized_ids: []}}
```

### After this PR:

The aggreate permission is the union of the parent workers. In this case, **all the composite workers** have process permissions.

```
aggregate: {permissions: {public: false, authorized_ids: {"node_1", ..., "node_N"}}}
```

## Why the fix is needed

There has historically been a hack that **granted open permissions for ALL aggregate tuples**. Here is the old incorrect code:

https://github.com/SubstraFoundation/substra-chaincode/blob/f2ba798baf8a9bd86b7973ccc4a038a12a90d678/chaincode/tuple.go#L217-L233

This hack **was fixed** with https://github.com/SubstraFoundation/substra-chaincode/pull/139. Since that PR was merged, aggregate tuples were correctly enforcing permissions. 

However, the enforced permissions were calculated incorrectly. This is what this PR fixes.

### DIFF
--- a/chaincode/tuple_aggregate.go
+++ b/chaincode/tuple_aggregate.go
@@ -66,7 +66,7 @@ func (tuple *Aggregatetuple) SetFromParents(db *LedgerDB, inModels []string) err
 	inModelKeys := tuple.InModelKeys
 	owner, err := GetTxCreator(db.cc)
 	if err != nil {
-		return errors.BadRequest(err, "could not transaction creator")
+		return errors.BadRequest(err, "could not get transaction creator")
 	}
 
 	authorizedIdsMap := map[string]bool{
--- a/chaincode/tuple_aggregate.go
+++ b/chaincode/tuple_aggregate.go
@@ -64,7 +64,15 @@ func (tuple *Aggregatetuple) SetFromInput(db *LedgerDB, inp inputAggregatetuple)
 func (tuple *Aggregatetuple) SetFromParents(db *LedgerDB, inModels []string) error {
 	var parentStatuses []string
 	inModelKeys := tuple.InModelKeys
-	permissions, err := NewPermissions(db, OpenPermissions)
+	owner, err := GetTxCreator(db.cc)
+	if err != nil {
+		return errors.BadRequest(err, "could not transaction creator")
+	}
+
+	authorizedIdsMap := map[string]bool{
+		owner: true,
+	}
+
 	if err != nil {
 		return errors.BadRequest(err, "could not generate open permissions")
 	}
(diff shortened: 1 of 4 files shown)


chaincode/tuple_aggregate.go
```diff
@@ -64,7 +64,15 @@ func (tuple *Aggregatetuple) SetFromInput(db *LedgerDB, inp inputAggregatetuple)
 func (tuple *Aggregatetuple) SetFromParents(db *LedgerDB, inModels []string) error {
 	var parentStatuses []string
 	inModelKeys := tuple.InModelKeys
-	permissions, err := NewPermissions(db, OpenPermissions)
+	owner, err := GetTxCreator(db.cc)
+	if err != nil {
+		return errors.BadRequest(err, "could not transaction creator")
```
mblottiere:
```suggestion
		return errors.BadRequest(err, "could not get transaction creator")
```

chaincode/tuple_aggregate.go
```diff
@@ -64,7 +64,15 @@ func (tuple *Aggregatetuple) SetFromInput(db *LedgerDB, inp inputAggregatetuple)
...
+	}
+
+	authorizedIdsMap := map[string]bool{
```
mblottiere: You don't seem to use the flag, only the key:
```suggestion
	authorizedIds := []string{}
```

AurelienGasser: I'm using a map here to find _unique_ msp ids. I wish there was a less verbose/convoluted way to do this in Go, but there isn't!

mblottiere: Ha good point 👍 You can then use a `map[string]struct{}` to lower the memory footprint. (this is yak shaving, feel free to dismiss this comment)

inelgnu:
does the current code handle the union of 
public: false with public: true? what is the output?

AurelienGasser:
@inalgnu the output always has public=False, regardless of parents

inelgnu:
> @inalgnu the output always has public=False, regardless of parents

This should not be the case.

The **union** between public:true and public: false should be equal to public: true.

Merging (public: True, authorized: []) with (public: False, authorized: [A,B]) should be public: True, authorized: [] in case of an aggregate.

AurelienGasser:
@inalgnu Thanks for your comment. I implemented permissions as the union of parent permissions.

Please note that, in order to be consistent with how we handle permissions in other cases, the **creator of the transaction** also gets added to the list of authorized ids (see [NewPermissions](https://github.com/SubstraFoundation/substra-chaincode/blob/beacebdd911d5e0bceb8655251206a1ffd78189f/chaincode/permissions.go#L62)). I don't think that's a behavior we want in the long run, since IMO giving permission should be explicit: we might want to create assets without granting permission to ourselves. cc @mblottiere

inelgnu (approved):

